PR #7, opened 2018-08-24
Using Deployment rather than StatefulSet for adminserver, registry and chartmuseum

ywk253100:
The components: adminserver, registry and chartmuseum are all stateless applications, so the Deployment should be used in the chart.


ywk253100:
@lanmingle The `subPath` makes sense. Could you submit your commits back to the repository after this PR merged?

mumrau:
I'm coming back here since no one replied on #260 .
I keep believing this change is a huge drawback as it cuts off totally automated update and upgrade path.
These are not stateless applicatinos at all since they use a PVC, which by nature, storage, well, stateful.

PR #9, opened 2018-08-28
Separate the ingress rule host and "externalDomain" in values.yaml

ywk253100:
The externalDomain is used to generate the ingress rule, but in the scenario that another proxy is deployed before Harbor, the externalDomain should be configured as the domain of external proxy and it should not be used to generate the ingress rule.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -30,28 +30,12 @@ release: {{ .Release.Name }}
 app: "{{ template "harbor.name" . }}"
 {{- end -}}
 
-{{- define "harbor.externalURL" -}}
-{{- if .Values.externalPort -}}
-{{- printf "%s://%s:%s" .Values.externalProtocol .Values.externalDomain (toString .Values.externalPort) -}}
-{{- else -}}
-{{- printf "%s://%s" .Values.externalProtocol .Values.externalDomain -}}
-{{- end -}}
-{{- end -}}
-
-{{/*
-Use *.domain.com as the Common Name in the certificate,
-so it can match Harbor service FQDN and Notary service FQDN.
-*/}}
-{{- define "harbor.certCommonName" -}}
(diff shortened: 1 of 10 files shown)


templates/_helpers.tpl
```diff
@@ -30,28 +30,12 @@ release: {{ .Release.Name }}
...
-{{- $list := prepend (rest $list) "*" -}}
-{{- $cn := join "." $list -}}
-{{- printf "%s" $cn -}}
-{{- end -}}
-
-{{/* The external FQDN of Notary server. */}}
-{{- define "harbor.notaryFQDN" -}}
-{{- printf "notary-%s" .Values.externalDomain -}}
+{{- define "harbor.isAutoGenedCertNeeded" -}}
+  {{- if and (and .Values.ingress.enabled .Values.ingress.tls.enabled) (not .Values.ingress.tls.secretName) -}}
```
reasonerjt: how about `and not .Values.externalURL`?
As the AutoGenCert is a shortcut to help user quickly deploy the chart.

templates/_helpers.tpl
```diff
@@ -30,28 +30,12 @@ release: {{ .Release.Name }}
...
-{{- $list := splitList "." .Values.externalDomain -}}
-{{- $list := prepend (rest $list) "*" -}}
-{{- $cn := join "." $list -}}
-{{- printf "%s" $cn -}}
-{{- end -}}
-
-{{/* The external FQDN of Notary server. */}}
-{{- define "harbor.notaryFQDN" -}}
-{{- printf "notary-%s" .Values.externalDomain -}}
+{{- define "harbor.isAutoGenedCertNeeded" -}}
```
reasonerjt: How'bout rename to `autoGenCert`?  I think this self-explains it is a flag itself.

PR #14, opened 2018-09-04
Add support to NodePort service in chart

ywk253100:
NodePort service can be used in the Kubernetes environment where the ingress controller is not installed


values.yaml
```diff
@@ -28,6 +28,30 @@ ingress:
     # replace the variable reference if you want to use your own TLS secret for notary
     notarySecretName: *tls_secretName
 
+# Setting ingress.enabled to false will enable the service, and
+# the nginx will be deployed as the proxy
+service:
```
reasonerjt: The term `service` is quite confusing, it's not very clear that this is the service that points to the nginx pod if nginx pod is to be deployed.
Or we better clarify it in the comments?

PR #27, opened 2018-09-11
correct order of 'and' operator

greenorchid:
the and operator was not taking affect for the deployment charts.


ywk253100:
@greenorchid Thanks for the commit!

PR #39, opened 2018-09-18
Create separate ingress resource for notary

jw-s:
Currently this helm chart only creates one ingress resource and shares the same TLS secret. 
This splits the two services into their own ingress resource and allows users to use a different secret.

To expect the same behavior, the default is to use the same value as what is set in `ingress.tls.secretName`

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -22,10 +22,11 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-body-size: "0"
   tls:
     enabled: true
-    # Fill the secretName if you want to use the certificate of 
-    # yourself when Harbor serves with HTTPS. A certificate will 
-    # be generated automatically by the chart if leave it empty
-    secretName: ""
+    # Fill in the empty string if you want to use your own TLS secret
+    secretName: &tls_secretName ""
+    # notary secret by default uses the value of secretName
+    # replace the variable reference if you want to use your own TLS secret for notary
+    notarySecretName: *tls_secretName
 
 harborAdminPassword: Harbor12345
 # The secret key used for encryption. Must be a string of 16 chars.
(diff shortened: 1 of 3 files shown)


ywk253100:
@jw-s Why we need two ingress rules? What's your use case?

jw-s:
@ywk253100 In our company wildcards are discouraged and therefore for this setup to work we require two TLS certs, one for the registry and the other for notary. Since ingress objects don't support multiple secrets. This was the solution.

ywk253100:
@jw-s We can use one ingress rule with multiple tls config, refter to https://github.com/kubernetes/ingress-nginx/blob/master/docs/examples/multi-tls/multi-tls.yaml.

values.yaml
```diff
@@ -284,6 +284,8 @@ redis:
 
 notary:
   enabled: true
+  tls:
```
ywk253100: I think we should move the `tls` to the `ingress` section with a different name

jw-s: @ywk253100 I never knew you could put multiple TLS config! Great, i'll follow this approach.

jw-s: @ywk253100 This has been changed, ready for review

values.yaml
```diff
@@ -25,7 +25,8 @@ ingress:
     # Fill the secretName if you want to use the certificate of 
     # yourself when Harbor serves with HTTPS. A certificate will 
     # be generated automatically by the chart if leave it empty
-    secretName: ""
+    secretName: &tls_secretName ""
+    notarySecretName: *tls_secretName
```
ywk253100: @jw-s Could you add some comment here to explain how to use these two config items?

jw-s: Done

Issue #70, opened 2018-10-26
LDAP and Email settings are not available in the chart


pierredadt:
From an operator and system administrator point of view, being able to deploy a product from scratch and 100% automated is a must have.
Choosing to remove the "users" settings from the deployment chart forces an otherwise 100% automated flow to be finished manually.
This choice hence violates the kubernetes paradigm aiming at making applications deployments fully automated and reproductible.

ywk253100:
The "user settings" should be set by API.

neith00:
I can understand that user settings should be set by API. But in this case I can't see a smooth deprecation process. We are also using configuration through env vars for LDAP and MAIL. As far from today I wasn't warn of a deprecation nor informed that there is an on going process to define user settings through api. (By the way I don't find anything about the configuration of the ldap server and certs in the swagger) It would have been smoother to plan the deprecation to let operators the time to update their deployments by offering both ways of configuration during a defined period.
Would you give us reference where to find email server and ldap server configuration in the API?

ywk253100:
Some refactoring works are in progress as described in https://github.com/goharbor/community/pull/12. After that, the `adminserver` will be removed and the environment variables that set user settings will not work anymore.

ywk253100:
Closing this one, feel free to reopen it if you have any further question.

darend:
@ywk253100 Will it still be possible to configure LDAP through the `harbor.cfg` file?

If so, another approach that allows for an automated chart installation is to separate the LDAP config from `harbor.cfg` into its own file. Users of the chart could then inject in the ldap config file through a volume mount.

Example: https://github.com/helm/charts/blob/798226aac49d8f28bca9b6afb03a89df8ce1d467/stable/datadog/values.yaml#L186-L198

ywk253100:
The settings will be removed from `harbor.cfg`

jimangel:
> the environment variables that set user settings will not work anymore.

@ywk253100 in [#71](https://github.com/goharbor/harbor-helm/pull/71#issuecomment-436899969) you said the reasons why email and ldap can't be configured via helm is in this PR (#71). However, the only relevant information I could find is the community proposal (not merged, blocked by @steven-zou) that states why they want to refactor the admin server. The article doesn't explain why env vars will not work anymore.

[...]

PR #71, opened 2018-10-26
Adding back email and ldap setting

pierredadt:
Hello,

From an operator and system administrator point of view, being able to deploy a product from scratch and 100% automated is a must have.
Choosing to remove the "users" settings from the deployment chart forces an otherwise 100% automated flow to be finished manually.
This choice hence violates the kubernetes paradigm aiming at making applications deployments fully automated and reproducible.

The PR aims at giving the opportunity, if chosen, to be able to deploy Harbor with Email and LDAP connections pre-configured.

What's new: 
- Added back Email settings, and a trigger on the `email.enabled` boolean value to activate of deactivate email configuration
- Added back LDAP settings, and a trigger on the `auth.mode` value to activate of deactivate LDAP configuration
- Added `ingress.labels` value to be able to customize the ingress in case of specific ingress classes
- Completed README.md to reflect above changes and document them

What can be improved:
- The LDAP groups settings are not set, because there is no documentation on the associated env. vars. names and default values. It might be interesting to complete this section


ywk253100:
@pierredadt Thank you for your PR, but as we discussed in #70, the user settings should be set by API.
We'll not accept this commit. Thanks again.
About the ingress label, can you explain it in more detail?

pierredadt:
@ywk253100 thanks for your input.

I get there is an alternative way to set said settings, and I could stand behind your choice with a more in-depth explanation. I still believe we should at least have the choice between the 2 solutions, especially since there is no real support/documentation on how to operate with API (except the swagger file in harbor repo).

Concerning ingress labels, certain ingress controllers can handle different ingress classes (e.g. http/https, network restricted, ...) and know what to implement depending on labels found in the manifest. 
The labels given are not necessarily the ones used by said controllers, and it is essential to let operators choose the options they need.

That said, do you mind sharing your strategy on users settings handling and the reasons behind it ?

ywk253100:
@pierredadt The reason why the user settings cannot be set by environment variables is explained in #70.

And as I know, different ingress controllers have different annotations to customize their behavior. So what is the label used for? Is there any ingress controller using labels?

ywk253100:
Closing due to inactivity, feel free to reopen it.

PR #84, opened 2018-11-15
Change access mode of volumes

ywk253100:
Change the access mode of volumes for jobservice, registry and chartmuseum: ReadWriteOnce -> ReadWriteMany

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -122,7 +122,7 @@ jobservice:
       # existingClaim: ""
       # storageClass: "-"
       subPath: "jobservice"
-      accessMode: ReadWriteOnce
+      accessMode: ReadWriteMany
       size: 1Gi
 # resources:
 #   requests:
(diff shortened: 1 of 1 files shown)


Timer:
What was the motivation behind this? Not all storage backends support this mode.

ywk253100:
@Timer The volumes of jobservice, registry and chartmuseum should be read/write by all replicas when deploying with multiple instances.

Timer:
I'm using a single replica right now and omitted this change.

However, this mode is incompatible with anyone deployed on vmware/vSphere: https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/faqs.html#is-readwritemany-volume-supported-with-vsphere-cloud-provider

ywk253100:
@Timer `RWM` works with a single replica even on vSphere but not work with multiple replicas, is this what you mean? If it is, this is the behavior we expect.

Timer:
If `RWM` modify works with a single replica, we should be good then. Thanks!

PR #102, opened 2018-11-29
Add checksum to deployments and statefulsets:

jchanam:
- Add checksum for deployments/statefulsets so they are redeployed if config change (https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change)
- Rename files to follow same naming


ywk253100:
@jchanam Thanks for the commit. However,  there is already a PR #96 to add the `checksum` to deployment/statefulset.

PR #104, opened 2018-12-05
Add support for customizing pod annotations

darend:
## what

Allow annotations to be added to pods

## why

We use [kiam](https://github.com/uswitch/kiam) to managing access to AWS, and it requires annotating pods with the role they should assume. In order to use S3 as a backend, we need this annotation added.

This should help with other use cases as well


ywk253100:
@darend Thanks for the PR.
I think the `nginx`, `database` and `redis` also need the annotation, right?
And add the property to the configuration list `in README.md`.

PR #108, opened 2018-12-06
Quote the chunksize to avoid the helm bug

ywk253100:
Due to the bug of helm https://github.com/helm/helm/issues/1707, the chunksize will be rendered in scientific notation. Quote it to avoid this.


ywk253100:
Use `--set-string` instead of `--set` to set `chunksize` when passing parameters via command line.

ywk253100:
Fixes https://github.com/goharbor/harbor/issues/6452.

Issue #119, opened 2018-12-14
helm upgrade error 0.3.0 to master


wu105:
On existing release using chart version 0.3.0, change to chart version master, run helm upgrade command, got the following (using helm v2.11.0):
```
Error: UPGRADE FAILED: PersistentVolumeClaim "ht-harbor-chartmuseum" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims && PersistentVolumeClaim "ht-harbor-registry" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims && StatefulSet.apps "ht-redis-master" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden. && StatefulSet.apps "ht-harbor-database" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden.
```

Rolling back to the failed revision generated a new "deployed" revision:
```
 helm history ht
REVISION	UPDATED                 	STATUS    	CHART       	DESCRIPTION                                                 
1       	Sat Nov 24 15:32:19 2018	SUPERSEDED	harbor-0.3.0	Install complete                                            
2       	Fri Dec 14 02:42:41 2018	SUPERSEDED	harbor-0.3.0	Upgrade complete                                            
3       	Fri Dec 14 02:59:12 2018	SUPERSEDED	harbor-0.3.0	Upgrade complete                                            
4       	Fri Dec 14 03:59:50 2018	FAILED    	harbor-dev  	Upgrade "ht" failed: PersistentVolumeClaim "ht-harbor-cha...
5       	Fri Dec 14 04:02:43 2018	DEPLOYED  	harbor-dev  	Rollback to 4 
```

Rolling back to revision 3 resulted in failure.  Then a upgrade using the same command line as revision 3 plus ---force succeeded, but a few pvs got recreated and the adminserver pod failed to initialized its database.
```
helm history ht
REVISION	UPDATED                 	STATUS    	CHART       	DESCRIPTION                                                 
1       	Sat Nov 24 15:32:19 2018	SUPERSEDED	harbor-0.3.0	Install complete                                            
2       	Fri Dec 14 02:42:41 2018	SUPERSEDED	harbor-0.3.0	Upgrade complete                                            
3       	Fri Dec 14 02:59:12 2018	SUPERSEDED	harbor-0.3.0	Upgrade complete                                            
[...]

PR #120, opened 2018-12-14
[0.3.0]Fix bug in the configmap of chartmuseum

ywk253100:
Pass the context "." as a param in calling "tempalte" function


ywk253100:
Fix bug reported in #119.

PR #126, opened 2018-12-18
Support for custom secret keys and certificates

### LINKED ISSUE #107: Registry Secret causes changes on every apply
## issue

Every time `helm diff` or `helm sync` is run changes are shown even if no config changes we made

## background

The registry secret generates a random password inline: https://github.com/goharbor/harbor-helm/blob/b5b422f91b8a574762ac980737a65fb2690bc98e/templates/registry/registry-secret.yaml#L9

Similarly with the notary certificates: https://github.com/goharbor/harbor-helm/blob/b5b422f91b8a574762ac980737a65fb2690bc98e/templates/notary/notary-cm.yaml#L11

fydrah:
This PR adds:
* Support custom CA and Certs for core and notary
  components, instead of generating new ones during
  install/upgrades
* Support custom secret keys for several components,
  instead of generating new ones during install/upgrades
* Fix chartmuseum secret dependency: if core secret
  changes, chartmuseum must be restarted

It also makes `helm upgrade` idempotent (no more secrets/certificates generation).

Fixes #107


ywk253100:
@fydrah Thanks for your PR.
Can you follow the same convention for the customized certificate as `expose.tls.secretName` and add more command to explain what the values are used?

fydrah:
Hi @ywk253100 ,

> Can you follow the same convention for the customized certificate as `expose.tls.secretName`

If I understand, each custom certificates/keys must be created by hand into a new secret, and included with a `secretName` style var.
I'm okay with it, but how should we manage [secret keys](https://github.com/fydrah/harbor-helm/blob/master/templates/core/core-secret.yaml#L11) ? Can we keep this method or should we also export them into an other secret ?

ywk253100:
I think we can keep your method for the `secret`.

And some other comments:
1) Can you add some comments to explain that if users want to avoid the issue #107, they should configure these items?
2) Update the comments of `certBundle` and `key` for `core`: the key pair is used to encrypt/decrypt the token generated by token service
3) Update the comments of `secret` for `core` and `jobservice`, they are used to do authentication when communicating.
4) Update the comments of `secret` for `registry`, refer to https://github.com/docker/distribution/blob/master/docs/configuration.md#http.

ywk253100:
Ping @fydrah

fydrah:
Hi, comments updated and use `secret` and `secretName` convention for var names.

Tested with the following override configuration:
```
core:
  secret: "ohb4AhchuaShai0i"
  secretName: "core-custom-secret"
jobservice:
  secret: "eip1aiLaeyua5Ahf"
registry:
  secret: "pop1ooNgielox6Ai"
chartmuseum:
  enabled: true
clair:
  enabled: true
notary:
  enabled: true
  secretName: "notary-custom-secret"
```
With helm deployment name `harbor`, the notary certificate `notary-custom-secret` has to be issued with `-subj "/CN=harbor-harbor-notary-signer"` alternative name.

ywk253100:
@fydrah Thanks!

Issue #10, opened 2018-08-28
Support node port service when no ingress controller is used


ywk253100:
Currently, the ingress controller is needed when deploying Harbor chart, but in some users' Kubernetes environment, the ingress controller is not installed at all. We need to provide the node port service to support those scenarios.

ywk253100:
Done by #14.

PR #129, opened 2018-12-27
Add support to gcs storage

ywk253100:
Provide the config for gcs properties to support storing images and charts


ywk253100:
Fix https://github.com/goharbor/harbor/issues/6558 and https://github.com/goharbor/harbor/issues/6448.

values.yaml
```diff
@@ -149,8 +149,8 @@ persistence:
       #realm: core.windows.net
     gcs:
       bucket: bucketname
-      # TODO: support the keyfile of gcs
-      #keyfile: /path/to/keyfile
+      # The base64 encoded json file which contains the key
+      keyfile: base64-encoded-json-key-file
```
reasonerjt: This is the key data rather than the key file, right?
Is it a convention to put the data in attribute `keyfile` when configuring gcs?

reasonerjt:
So sorry for the delayed review, please see my comment.

values.yaml
ywk253100: Yes, it is the content of the key file. And the chart of chartmusuem uses the same way to support gcs.

reasonerjt: So is it appropriate to call it `keyfile` here?  maybe I was misled by the removed comment, but it may make user think he should put the file path here.

EvertonSA:
Hi, sorry for raising this up, but it I tried different attempts and they all show the same result: 

> Error: release harbor failed: Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 0, error found in #10 byte of.....

I tried to pass the actual file, the file content, a base 64 encrypted file, and a base 64 encrypted file value. 

Am I doing something wrong?

EvertonSA:
> Hi, sorry for raising this up, but it I tried different attempts and they all show the same result:
> 
> > Error: release harbor failed: Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 0, error found in #10 byte of.....
> 
> I tried to pass the actual file, the file content, a base 64 encrypted file, and a base 64 encrypted file value.
> 
> Am I doing something wrong?

My bad, I was serving the Service Account GCS key instead of the GCS key.

PR #130, opened 2018-12-28
Fix s3 backend push retrying issue

toby1991:
When using `s3` for the backend type, `docker push` cannot successfully push the image, and will always be `retrying`.
If switch the backend to `filesystem`, this issue was gone.

Adding the following to the registry config.yml manually fixes my issue.

```yaml
storage:
  redirect:
    disable: true
```

And thanks for @rizwan-kh  helping!!

issue links:
https://github.com/goharbor/harbor/issues/5240
https://github.com/vmware/harbor-boshrelease/issues/15

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ The following table lists the configurable parameters of the Harbor chart and th
 |`persistence.persistentVolumeClaim.redis.subPath`     | The sub path used in the volume. If external Redis is used, the setting will be ignored | |
 |`persistence.persistentVolumeClaim.redis.accessMode`     | The access mode of the volume. If external Redis is used, the setting will be ignored | `ReadWriteOnce` |
 |`persistence.persistentVolumeClaim.redis.size`     | The size of the volume. If external Redis is used, the setting will be ignored | `1Gi` |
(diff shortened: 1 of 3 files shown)


ywk253100:
@toby1991 You need to find out the root cause of `docker push` failure. We cannot just disable the redirection as it is more effective in most cases.

We can make it configurable in `values.yaml` with the default value as enabled.

toby1991:
> @toby1991 You need to find out the root cause of `docker push` failure. We cannot just disable the redirection as it is more effective in most cases.
> 
> We can make it configurable in `values.yaml` with the default value as enabled.

I've made a push that make it configurable in `values.yaml`

ywk253100:
Thanks a lot. Some comments as below:
1. The `redirect` shouldn't be just a property for `s3` only, it can be applied to other storages, so put it as a property of `imageChartStorage` is more reasonable. 
2. Add some comments for the property in `values.yaml`
3. Add the property in the `README.md`
4. Squash all your commits into one

You can refer to https://github.com/docker/distribution/blob/master/docs/configuration.md#redirect for the usage of `redirect`.

ywk253100:
@toby1991 Thanks!

wangjb1803:
in my case ,the waf  integrate with alb block some requests, i solved it by adding a rule to allow harbor requests

PR #138, opened 2019-01-10
Fix download chart issue in V1.0.0

mmpei:
Signed-off-by: peimingming <peimingming@corp.netease.com>


ywk253100:
@mmpei Thanks a lot.

PR #140, opened 2019-01-11
Add checksum in chartmuseum to core secret

jchanam:
Chartmuseum depends on core secret, using the env `BASIC_AUTH_PASS` from `harbor.core` secret.

This change adds a checksum on the chartmuseum deployment to force the recreation of the pods, allowing them to get the new value for the env variable.


ywk253100:
@jchanam Thanks for the PR.
I think the `core`, `jobservice` and `registryctl` have the same issue, could you update the deployments of them either?

`core` depends on the secret of `jobservice`.
`jobservice` depends on the secret of `core`.
`registryctl`(in the same pod with `registry`) depends on the secret of both `core` and `jobservice`.

jchanam:
@ywk253100 done!

I've seen that checksums were missing also for notary. All added now.

PR #141, opened 2019-01-14
Fix pod resources

jchanam:
- Add ifs to resources in pods, to have a cleaner output
- Add core resources in deployment
- Add portal resources in deployment
- Separate registry resources for each component (registry/controller)
- Move commented resources for registry in values.yaml to match new separation
- Add missing notary commented resources in values.yaml


ywk253100:
@jchanam Thanks a lot! Could update the README.md?

PR #143, opened 2019-01-17
Allow user to specify custom URI for harbor

max-k:
I provide custom configuration keys meant to configure ingress to use custom URI.

It can be usefull if you want to use the same hostname for harbor and notary, for example. Or to re-use an  hostname where you have already deployed an application.

This merge request cannot be used on its own because api URLs are hardcoded in Angular application.

I'll open an issue on harbor project to ask for help on this part.

### DIFF
--- a/templates/ingress/ingress.yaml
+++ b/templates/ingress/ingress.yaml
@@ -35,27 +35,51 @@ spec:
   - host: {{ $ingress.hosts.core }}
     http:
       paths:
+      {{- if $ingress.paths.core }}
+      - path: {{ $ingress.paths.core }}
+      {{- else }}
       - path: /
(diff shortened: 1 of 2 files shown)


ywk253100:
@max-k Thanks for your commit.
As mentioned by @reasonerjt  in https://github.com/goharbor/harbor/issues/6772#issuecomment-455956758, the docker client doesn't support prefix. So maybe we can add a prefix only for notary service to make the core and notary serve on the same hostname. If it is ok, I think we can remove the `hosts.notary` in the values.yaml directly.

ywk253100:
Ping @max-k

ywk253100:
Closing as inactivity.

PR #159, opened 2019-02-18
Extract some fixed values in adminserver-cm.yaml to values.yaml

airadier:
In our current deployment, we are tweaking some values that hardcoded in adminserver-cm.yaml. This PR extracts this fixed values (AUTH_MODE, SELF_REGISTRATION, PROJECT_CREATION_RESTRICTION) and LDAP settings to values.yaml

This change could also be applied to master branch.

### DIFF
--- a/templates/adminserver/adminserver-cm.yaml
+++ b/templates/adminserver/adminserver-cm.yaml
@@ -41,9 +41,22 @@ data:
   # The following properties have no default value in Harbor core if not 
   # set via env. They should be removed after the refactoring:
   # AUTH_MODE, SELF_REGISTRATION, PROJECT_CREATION_RESTRICTION 
-  AUTH_MODE: "db_auth"	
-  SELF_REGISTRATION: "on"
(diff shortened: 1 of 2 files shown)


ywk253100:
@airadier Thanks for your PR. 
We have discussed the same issue in #70. The `adminserver` has been removed from Harbor and the user settings cannot be set via environment variables anymore, so the PR cannot be accepted.

airadier:
Nice to now that it will be removed. However for 1.0.0 branch and current version of Harbor, they are still there and working, aren't they? I guess the merge makes no sense for future versions, but for version 1.0.0 we require it, unless there is other approach for setting that values.

Thanks.

ywk253100:
As users cannot configure these items in the chart whose version > 1.0.0 by editing values.yaml , why do they need to configure them in version 1.0.0 using this way? They should use the compatible one instead. 

And as the version 1.0.0 has been released, only bug fixes can be accepted into the branch.

airadier:
Sorry, you are right, I noticed that those values have no effect in 1.0.0.

Issue #48, opened 2018-09-20
lost commit history during move to new repo


paulczar:
Hi,

I just found the new repo for harbor's helm repo and noticed the commit history has all been squashed out.   It would be really nice to have used `git filter-branch --prune-empty --subdirectory-filter FOLDER-NAME  BRANCH-NAME` to split it out while keeping the commit history.

Also I believe this also makes the DCO for the original commit somewhat contentious as it does not list all contributors of code to the commit.

There is probably not much that can be done, its possible we could go back to the original repo and git filter-branch it and then re-apply the commits in here.  I see there's a [incomplete] list of contributors in a contributors file so that at least helps acknowledge the hard work done by the various contributors.

scottrigby:
I'm a strong advocate of retaining git history when moving or splitting repos. I had started on a tool to automatically do this for the Helm's upcoming [Distributed Search Hub](https://github.com/helm/community/blob/master/proposals/distributed-search.md), but haven't completed that as a full project yet. It likely wouldn't help this project anyway, since it would take some manual work to get it back to the current state but with full history at the start.

It may be slightly tricky to splice history in at the first or second commit, but you should be able to do that, resolve conflicts, then replay the newer commits on top. If you want to create a test branch and want another set of eyes on it let me know.

I'm not sure about the DCO question on original commits, but that may be something worth looking into.

reasonerjt:
@paulczar 
Thanks for pointing it out.
Sorry we didn't consider that while splitting the repo.

We'll make update to contributors list to acknowledge the contribution by community, and investigate how we can restore the commit histories later.

scottrigby:
OK I spliced it back in. I can't open a PR because I changed the initial commit (added history from before chart was deleted from https://github.com/goharbor/harbor), then replayed all of this repo's commits on top. But you can - if you want to - take the work I did in my branch (https://github.com/scottrigby/harbor-helm/tree/recover-history), and force push it to this repo.

You can also redo this yourself (the only file that requires a manual resolution is `.gitignore`).

### My steps

1. Check out new repo:
    ```
    WORKSPACE=~/development/github.com/
    DIR=goharbor/harbor-helm
    URL=git@github.com:goharbor/harbor-helm.git
    cd $WORKSPACE && mkdir -p $DIR && git clone -o new  $URL $DIR && cd $DIR
    ```

1. Add old repo as remote:
    ```
    URL=git@github.com:goharbor/harbor.git
    git remote add old  $URL
    ```

[...]

PR #160, opened 2019-02-18
Ingress requires services to be NodePort not ClusterIP

paulczar:
updates service logic so that if ingress is set the services are exposed
as nodeports where needed.


ywk253100:
@paulczar Why do we need this PR?

paulczar:
Ahh this may actually be heavy handed.  I'm using Google ingress controller which requires type to be nodeport, I bet a in-cluster ingress might be fine with clusterip.  I'll update this with smarter logic.

paulczar:
GCE ingress is external to the cluster so the services being exposed via ingress need to be NodePort.

Also GCE ingress path rules are very specific, so we need to loosen them with a wildcard.

This leave default behavior as is.

paulczar:
@ywk253100 does my followup comment and commit clear my intentions up for you ?

ywk253100:
@paulczar Please give me more time to investigate the GCE ingress. Thanks.

ywk253100:
@paulczar All your PRs are closed automatically when we fixing #48, you need to re-do the works in the PRs and commit them again, sorry for that.

PR #165, opened 2019-02-20
Reverts #150

ywk253100:
Reverts #150
The PR #150 causes new issue: [goharbor/harbor#6903](https://github.com/goharbor/harbor/issues/6903).
Harbor doesn't support conifgure chart_url fow now, there is an issue [goharbor/harbor#6572](https://github.com/goharbor/harbor/issues/6572) to track this. After the issue is fixed, we can add the support in chart

### DIFF
--- a/templates/chartmuseum/chartmuseum-cm.yaml
+++ b/templates/chartmuseum/chartmuseum-cm.yaml
@@ -22,7 +22,7 @@ data:
   DISABLE_API: "false"
   DISABLE_STATEFILES: "false"
   ALLOW_OVERWRITE: "true"
-  CHART_URL: {{ .Values.externalURL }}/chartrepo
+  #CHART_URL: {{ .Values.externalURL }}/chartrepo
   AUTH_ANONYMOUS_GET: "false"
   TLS_CERT:
   TLS_KEY:


darend:
@ywk253100 https://github.com/goharbor/harbor/pull/7719 was merged and addresses https://github.com/goharbor/harbor/issues/6572

Does that mean the helm chart can now be updated to support absolute URLs?

ywk253100:
@darend Yes

darend:
I’d be happy to help. It’s not clear to me how to make use of the change that was introduced to chart server. Is it as simple as setting the chart url like we previously attempted?

ywk253100:
@darend Yes, just need to set the environment variable `CHART_URL ` for chartmuseum. And I think we should expose the configuration item in `values.yaml` either.

darend:
> @darend Yes, just need to set the environment variable CHART_URL for chartmuseum. And I think we should expose the configuration item in values.yaml either.

If we are add support for setting `CHART_URL` as the absolute url we will need to be mindful of backwards support. We previously saw issues where chart museum did not support changing the URL (https://github.com/helm/chartmuseum/issues/157). The chart should make use of a flag indicating if `CHART_URL` should be absolute or relative. Perhaps defaulting to the old, relative, behavior.

If this sounds good, I can submit a fresh PR

ywk253100:
Using the relative url as the default value makes sense, please submit your PR, thanks.

Issue #155, opened 2019-02-13
helm templates currently do not support limits.cpu,limits.memory,requests.cpu,requests.memory


jimangel:
I am happy to PR this functionality in, but does anyone know the recommended settings or minimum requirements for each component?

ywk253100:
@jimangel There is a configuration item `resources` for each components in `values.yaml`, it contains `requests` but no `limits` and I find that the configuration doesn't apply to some components, such as `core`. 

It'll be appreciated if you can contribute the `limits` part and I think these configuration can be commented by default.

ywk253100:
@jimangel  The PR #172 is closed automatically when we fixing #48, you need to re-do the works in the PR and commit them again if you would like, sorry for that.

PR #172, opened 2019-02-28
adding limits

jimangel:
Fixes: https://github.com/goharbor/harbor-helm/issues/155


jimangel:
If you want an easy way to test / run this against values.yaml and generate the template:
```bash
sed -i 's/#resources:/resources:/g' values.yaml
sed -i 's/# requests:/ requests:/g' values.yaml
sed -i 's/# limits:/ limits:/g' values.yaml
sed -i 's/#   memory:/   memory:/g' values.yaml
sed -i 's/#   cpu:/   cpu:/g' values.yaml
```

PR #183, opened 2019-03-11
update ingress to support GCE Ingress Controller

paulczar:
GCE ingress controller is outside the cluster and therefore needs
the services to be exposed via a nodeport.

GCE ingress requires wildcards for the ingress path rules, otherwise
it does exact matches.


reasonerjt:
@paulczar 
Thanks a lot for this PR.
However, we currently don't verify GCE on our pipeline,  I assume you have verified it on your env, right?

paulczar:
Sure have!   The key for it working is:

* services set to NodePort ( because the gcp lb is external to the cluster )
* paths have `*` at the end as gcp lb is strict

PR #184, opened 2019-03-11
Add support for UAA

paulczar:
This allows a user to configure harbor to use UAA for authentication.


ywk253100:
@paulczar The authentication related settings cannot be set via environment variables anymore after removing the `adminserver`.

ywk253100:
Will closing the PR.

PR #193, opened 2019-03-18
Set default values for TLS_CERT and TLS_KEY in the Chartmuseum ConfigMap

### LINKED ISSUE #192: TLS_CERT and TLS_KEY are unspecified in Chartmuseum configmap
The values for `TLS_CERT` and `TLS_KEY` in the Chartmuseum configmap are null, which means that generating the ConfigMap and trying to apply it using `helm template` fails since the ConfigMap is invalid.

The fix is straightforward: set the values to `""`, which I'll do in a PR. I'm not sure if there's a reason for these values to even be present though since I don't see where you can ever specify values for these keys.

protochron:
This sets proper default values for TLS_CERT and TLS_KEY in the
Chartmuseum ConfigMap.

Closes #192

Signed-off-by: Dan Norris <dan.norris@netapp.com>


sdake:
@reasonerjt what is needed to get this PR moving?

Cheers
-steve

Hokwang:
@ywk253100 when will you release patch version with this PR ? I think no one can use chartmuseum now.

Hokwang:
@ywk253100 Could you tell me your plan for next helm chart release ?

Hokwang:
@ywk253100 I think there's no workaround to deploy chartmuseum using helm chart v1.1.0 , right? please reply. I am looking forward to releasing this PR.

ywk253100:
@Hokwang  Refer to the #192 to find the workaround

Hokwang:
@ywk253100 Many thanks, I can handle, I deployed chartmuseum successfully now.

By the way, do you have the below release policy ?
1. if harbor release new version, then harbor-helm release new version.
2. if there's a problem in chart, wait harbor's new release(patch), do not patch harbor-helm only.

I think you can release harbor-helm v1.1.1 as you already knew, and that is good way for communicate with harbor users.

ywk253100:
@Hokwang Yes, if there is new Harbor release or bug found in chart, we'll release a new version for Harbor chart. And the reason why we don't release a new one for #192 I have explained it in the comment of #192.

PR #195, opened 2019-03-19
Allow ability to disable self-registration

paroxp:
## What

In our use case, we'd like not to allow people to register into our
portal, and would prefer to disable that option declaratively via
`values.yaml` file.

I'm not certain that this will be picked up by harbor application but
previously similar thing has been part of the AdminServer. Now that it's
gone there is no trace of using this variable.

Re-adding it and allowing a customisation would be great.

## How to review

- Code review may be sufficient
- I'd like a confirmation weather or not that would be picked up by the app itself


ywk253100:
@paroxp The self-registration cannot be set via environment variable anymore. You need to configure it on the portal or through system setting API.

ywk253100:
Will closing the PR.

PR #198, opened 2019-03-21
Place the redis password in the registry secret not the configmap.

josecv:
The Redis password belongs with other passwords, in the registry secret, instead of in the registry configmap.


ywk253100:
@josecv Thanks for the PR.

PR #199, opened 2019-03-22
Allow the chart to use pre-defined secrets instead of generating them.

### LINKED ISSUE #189: Chart should allow specifying existing secrets
Right now, this helm chart generates all the secrets, unconditionally, from values in the `values.yaml`. For those of us who like to commit our `values.yaml`s this is not ideal, as it requires that we commit our passwords, for example our S3 access key.

It seems like the best practice is to allow the user to supply an `existingSecret` (e.g. in the [postgres chart](https://github.com/helm/charts/blob/master/stable/postgresql/values.yaml#L90)).

Something similar could be done here -- for example for the registry, if `registry.existingSecret` is set, don't generate a secret and just trust the user to figure out what the secret should look like on their own (which allows them to self-manage the lifecycle of their passwords).

If there's interest, I can find some time to do this PR.

josecv:
Closes #189

Signed-off-by: jose <jose@autonomic.ai>


bmcstdio:
@josecv do you plan on resuming work on this? 🙏

josecv:
@bmcstdio i will rebase the branch to get rid of the conflicts, but if there's new requirements that need to be captured by the PR i'd rather hand it off at that point :)

@ywk253100 wdyt? does this look good (merge conflicts aside)?

ywk253100:
@josecv Let's discuss this in issue #189

cnmcavoy:
I have a rebased copy of this PR in https://github.com/cnmcavoy/harbor-helm/tree/existing-secrets.

sathieu:
> I have a rebased copy of this PR in https://github.com/cnmcavoy/harbor-helm/tree/existing-secrets.

@cnmcavoy Please create a PR for this :pray:.

florentinadolf:
This issue is being solved in #1179

PR #202, opened 2019-03-22
Allow mounting non-self signed certs.

josecv:
This is needed because usually when a certificate is not self-signed, and generated by [cert-manager](https://github.com/jetstack/cert-manager) the tls secret will not contain a `ca.crt` field.

Without this, certificates generated by cert-manager can't be used with harbor. It seems as though mounting `tls.crt` will work if the certificate is rooted at a trusted CA, which cert-manager certs typically are.

Signed-off-by: jose <jose@autonomic.ai>


josecv:
I think different approaches are possible here, such as parameterizing `ca.crt` somewhere into `.Values` instead, so let me know if you'd like me to re-do this PR differently.

paulczar:
In my experience, certmanager still creates a ca.crt key in the secret, but gives it a nil value.  Won't that render an empty file?

Doublemine:
> In my experience, certmanager still creates a ca.crt key in the secret, but gives it a nil value. Won't that render an empty file?

@paulczar  What version of cert-manager are you using? In my test, on version of `v0.5 and below` cert-manager `ca.crt` key not in the secret.

ywk253100:
This should already be fixed in https://github.com/goharbor/harbor-helm/pull/348.

PR #205, opened 2019-03-25
Add a parameter to enable prometheus metrics from the registry.

josecv:
Docker registry can [quite easily](https://ops.tips/gists/retrieving-docker-registry-metrics-using-prometheus/) export prometheus metrics. This is quite nice, and should be easy to enable from the helm chart.


philandstuff:
We would also be very interested in this feature.  Other than rebasing this PR to resolve conflicts, is there anything that can help move it forward?

josecv:
@philandstuff I've rebased the branch :)

That said, I haven't played with harbor in some time, so it's possible this PR doesn't work. I suggest pulling it and testing it; if it works for your use case we can look into having the harbor team review it.

templates/registry/registry-dpl.yaml
```diff
@@ -21,6 +21,11 @@ spec:
         checksum/secret: {{ include (print $.Template.BasePath "/registry/registry-secret.yaml") . | sha256sum }}
         checksum/secret-jobservice: {{ include (print $.Template.BasePath "/jobservice/jobservice-secrets.yaml") . | sha256sum }}
         checksum/secret-core: {{ include (print $.Template.BasePath "/core/core-secret.yaml") . | sha256sum }}
+{{- if .Values.registry.enablePrometheus }}
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "5001"
+        prometheus.io/scrape: "true"
```
philandstuff: What do you expect to consume these annotations? I did some searching and it looks like `prometheus.io/scrape` used to be present in some example configurations, but it was removed in https://github.com/prometheus/prometheus/pull/4131 because people copied it literally rather than using it as inspiration.

(From our point of view, I expect to be using prometheus-operator's `ServiceMonitor` to configure what to scrape, so I'm pretty agnostic to these annotations).

philandstuff:
Thanks for the rebase! It looks fine to me (apart from one thing, comment above).

templates/registry/registry-dpl.yaml
josecv: hmm, good point -- I must have copied these from those examples. I'll remove the annotations from the PR as different configurations will configure scraping in different ways.

karl-johan-grahn:
@philandstuff I think @josecv fixed your comment, is this ready to go now?

philandstuff:
Sure it is. But I’m not a maintainer so I can’t really do much here.

mumrau:
I would like to make this issue alive, this looks like a promising PR.
I wonder if we could include the `ServiceMonitor` resource as well, which combines with `prometheus-operator` which is very widely used now. Of course it would be disabled by default to avoid clutter.

josecv:
> I wonder if we could include the ServiceMonitor resource as well, which combines with prometheus-operator which is very widely used now

personally i prefer it when charts don't bind themselves to particular cluster components (such as the prometheus operator), so imho the `ServiceMonitor` doesn't really belong here

That said, it's up to the maintainers whether they want that added in, ofc. although in that case it should probably be in a separate PR as i don't really have the `prometheus-operator` knowledge to implement it :)

josecv:
@ywk253100 do you have any thoughts on this PR? :)

ywk253100:
Sorry for the late reply, I'll review the PR and try to make it forward.

ywk253100:
Closing as Harbor chart has already support the metrics

PR #210, opened 2019-04-03
Recombines Ingress for Core and Notary- Fixes #206

### LINKED ISSUE #206: #185 broke some ingress controllers
Most ingress controllers do not support having multiple ingress with the same host rule and will create seperate loadbalancers for each ingress each with their own IP address which makes it impossible to do both.   We need to either keep them in the same ingress rule, or we need different hostnames.

tybritten:
Non-nginx ingress controllers can't handle multiple ingresses with the same host/ip- this recombines the notary ingress back into the main ingress template. 
Fixes #206 

Signed-off-by: Tyler Britten tyler@britten.us
Signed-off-by: Paul Czarkowski username.taken@gmail.com


ywk253100:
@tybritten @paulczar Thanks for the PR. But we need the `rewrite` annotation here https://github.com/goharbor/harbor-helm/blob/master/templates/ingress/notary-ingress.yaml#L12 to rewrite the `/notary` request to `/`.
We cannot just rewrite one path if we put the two ingresses in a single file. But if we put them in two files, there is the issue described in #206. And if we keep the same way as previous version of Harbor chart we need to configure two domain names(one for core service, the other one is for notary) for Harbor although Harbor is just one registry service.

ywk253100:
Any suggestion will be appreciated.

paulczar:
It looks like the new nginx rewrite rules that use capture groups may help here ( https://kubernetes.github.io/ingress-nginx/examples/rewrite/ )  as it would only rewrite rules that have a capture group... but that's still nginx specific.   

I would suggest the best way forward is to have a different hostname for notary.

tybritten:
I agree on separate host names. It's common for other projects like spinnaker (spinner.example.com, gate.spinnaker.example.com) and is the most compatible. If we agree, I update the PR

tybritten:
should we basically just revert this commit? https://github.com/goharbor/harbor-helm/pull/180/commits/e026825ae4c9bc06d1fee92a90476405eac2d2b6

ywk253100:
@tybritten @paulczar I did some investigation and found that the both `traefik` and `nginxinc/kubernetes-ingress` support ingress merging strategy:
 https://docs.traefik.io/user-guide/kubernetes/#multiple-ingress-definitions-for-the-same-host-or-hostpath
https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/mergeable-ingress-types
What kind of ingress controller are you using?

tybritten:
@ywk253100 Any of the controllers that manipulate an external LB for ingress (GCE, AWS, etc)- they don't support url rewrites.

paulczar:
We should revert the change and have it fall back to separate hostnames,  its not ideal, but it is the most compatible option across ingress controllers.   I'm not sure VMWare's own NSX-T ingress controller will even support this merging of ingress host rules.

ywk253100:
We got a requirement that the Harbor chart should support configuring no hosts so that all requests sent to the ingress controller will be sent to Harbor service. In this case, users don't have to get the domain names that can be resolved by DNS to configure Harbor, they can just use the domain name of the ingress controller.

So if we use two hostnames, the requirement will not be met. 

Maybe we can add the Nginx before other components even though we expose the service via ingress.

ywk253100:
The Nginx is responsible to rewrite the requests.

paulczar:
Yeah if its critical for the same hostname (or no hostname) in requirements, then I think using nginx in front of the other apps makes sense ...  would also allow the end user to choose between a service of type loadbalancer or an ingress controller (would only be needed if user wanted an ingress controller provided SSL) since nginx would be doing the L7 routing work.

ywk253100:
PR #225 merged, closing this one.

Issue #188, opened 2019-03-15
Chartmuseum PVC is created even if chartmuseum is disabled


arcalys:
Hi,

Just ran a few tests with this chart, and it seems like the chartmuseum PVC is created even though chartmuseum.enabled is set to false in values.yaml.

For now, I added the following to chartmuseum-pvc.yaml :
```yaml
{{- if .Values.chartmuseum.enabled }}
...
{{- end }}
```

Issue #111, opened 2018-12-07
Database password is stored in Configmap as clear text


darend:
The password for the clair and notary databases are stored in a Configmap as clear text. It should be stored as Secret for security.

A template is used to construct the postgres URL with the password:
https://github.com/goharbor/harbor-helm/blob/c263ec91c71067565317bf08aa212989d6a156ca/templates/_helpers.tpl#L133-L143

Its then used in the configmaps:
https://github.com/goharbor/harbor-helm/blob/c263ec91c71067565317bf08aa212989d6a156ca/templates/clair/clair-cm.yaml#L15
https://github.com/goharbor/harbor-helm/blob/c263ec91c71067565317bf08aa212989d6a156ca/templates/notary/notary-cm.yaml#L35
https://github.com/goharbor/harbor-helm/blob/c263ec91c71067565317bf08aa212989d6a156ca/templates/notary/notary-cm.yaml#L59

draeron:
this is critical security issue, on our cluster, all pod's stdout get collected and sent to a elasticsearch cluster which means the value of our password get indexed.

ywk253100:
@darend Thanks for your feedback, we'll fix it.

ywk253100:
@darend The issue of password leaking is tracked by https://github.com/goharbor/harbor/issues/7510.

darend:
> The issue of password leaking is tracked by goharbor/harbor#7510.

@ywk253100 Thanks. That issue tracks logging of passwords, there is still the issue that the password is stored in a ConfigMap vs Secret

PR #218, opened 2019-04-18
Fix #181 #188 #111

### LINKED ISSUE #181: The certificate for token service shouldn 't be generated every installation
The chart has the same issue with https://github.com/goharbor/harbor/issues/6824.

ywk253100:
Fix #181 #188 #111


reasonerjt:
Hi I personally have some concern to include the fix to #111  by storing configmap in secret, this works but may not be a good solution.  could you split the PR?  Other change seems good to me.

PR #223, opened 2019-04-30
Provide a default key pairs for signing the token

ywk253100:
The certificate/private key pair is needed for robot account and cannnot regenerated everytime install the chart, so this commit provides a default ones


ywk253100:
Fixes #181, #188

PR #239, opened 2019-05-23
Fix #192 - empty variables in chartmuseum/chartmuseum-cm.yaml

### LINKED ISSUE #192: TLS_CERT and TLS_KEY are unspecified in Chartmuseum configmap

ruzickap:
Fix #192 - empty variables in chartmuseum/chartmuseum-cm.yaml


ywk253100:
@ruzickap PR #193 is merged to fix #192, closing this one, thanks anyway.

PR #241, opened 2019-05-27
Fix #152 Add imagePullSecrets to deployments with a global default

### LINKED ISSUE #152: Add imagePullSecrets and namespaces to helm template
You can see the example below for imagePullSecrets and the same could be done for namespaces.

```
spec:
      {{- if .Values.image.pullSecret -}}
      imagePullSecrets:
        - name: {{ .Values.image.pullSecret }}
      {{- end }}
```

This will help companies deploy harbor using helm + private internal repository and/or scope it to a namespace.

seyhbold:
This PR adds a global default:

```
imagePullSecrets: []
```

and an `imagePullSecrets:` for every deployment.
to `values.yaml`

The template first look in the specific entry in `values.yaml` and default to the global entry which is an empty array by default.

Fixes #152


values.yaml
```diff
@@ -224,6 +224,11 @@ persistence:
 
 imagePullPolicy: IfNotPresent
 
+# Use this set to assign a list of default pullSecrets
+imagePullSecrets: []
```
ywk253100: I think we can remove the `[]` at the end of this line as users need to do that if they want to add a secret

values.yaml
```diff
@@ -235,6 +240,9 @@ nginx:
   image:
     repository: goharbor/nginx-photon
     tag: dev
+    # Override default imagePullSecrets
```
ywk253100: Is it necessary to provide a secret for each image? In what scenario this is needed?

values.yaml
seyhbold: You're right. I'll remoe it.
Should I leave it as a comment?

values.yaml
seyhbold: I am using public registries like hub.docker.com or quay.io along with a locally hosted for my own developments.
I do not want to see the list of all secrets in every deployment, but this is more to my own taste.
It is definitely not necessary.
Do you want it removed?

values.yaml
seyhbold: Ah, I should have read more thoroughly. I'll remove the `[]`

values.yaml
ywk253100: I think that keep only one overall config is better.

ywk253100:
Add DCO is needed for you PR.

README.md
```diff
@@ -145,12 +145,14 @@ The following table lists the configurable parameters of the Harbor chart and th
...
 | `nginx.image.tag`                                                           | Image tag                                                                                                                                                                                                                                                                                                                                       | `dev`                           |
+| `nginx.image.imagePullSecrets`                                              | Add imagePullSecret names to the nginx deployment                                                                                                                                                                                                                                                                                               |                                 |
```
ywk253100: I think these items are not needed anymore either

seyhbold:
Uhh sorry, I forgot.

README.md
seyhbold: Uhh sorry, I forgot.

ywk253100:
@seyhbold Thanks

PR #242, opened 2019-05-30
when tls disabled. set ssl-redirect: "false" 

sunnoy:
when tls disabled , the annotations of the ingress object can make docker client "http: server gave HTTP response to HTTPS client" error, so need change


ywk253100:
You should add the IP or domain name(192.168.0.1:80) to the `--insecure-registries` option of docker. Note the `80` at the end of the IP to avoid the  error `http: server gave HTTP response to HTTPS client`.

Is this your case? Or can you explain what your PR fixes and how?

ywk253100:
Closing it as inactivity.

PR #244, opened 2019-05-31
Generated notary certificates in secret instead of configmap

### LINKED ISSUE #243: Generated notary certificate stored in configmap as clear text
The automatically generated notary certificate is stored in the notary-configmap in clear text. This should be in a secret.

This could also simplify the server-config.postgres.json and signer-config.postgres.json, since the generated could be mounted on the same path as the optionally user supplied secret, and there would be no reason to have logic around where to find the .crt and .key files.

I'm willing to give this a shot.

PaulFarver:
Fixes #243 
- Store automatically generated notary certificates in a k8s secret for better security
- Unify notary certificate mount templating, for reduced complexity in notary-cm.yaml, notary-server.yaml and notary-signer.yaml

### DIFF
--- a/templates/notary/notary-cm.yaml
+++ b/templates/notary/notary-cm.yaml
@@ -7,16 +7,6 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
     component: notary
 data:
-  {{ $ca := genCA "harbor-notary-ca" 365 }}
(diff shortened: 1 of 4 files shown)


ywk253100:
@PaulFarver Thanks for the PR!

PR #249, opened 2019-06-01
Add support for configuring external auth

paulczar:
The new configuration API is a bit unwieldy and makes it hard to
automate user configuration when starting with helm.  This allows
a user to choose what settings to set and Helm will display a
`curl` command at the end of installation for the user to run.

Also allows for using an external ca cert for uaa.

Ideally this would be done as a job `post` install via a harbor-cli
or similar, but this is a suitable start.


paulczar:
Here's the notes output given:

```
NOTES:
Please wait for several minutes for Harbor deployment to complete.
Then you should be able to visit the Harbor portal at https://core.harbor.pivlab.gcp.paulczar.wtf.
For more details, please visit https://github.com/goharbor/harbor.


If this is your first time Installing Harbor,
You may need to use the Harbor Configuration API to
finish your configuration by running:

$ curl -i -X PUT -u "admin:xxxxxxx" \
  -H "Content-Type: application/json" \
  https://core.XXX.wtf/api/configurations \
  -d '{"auth_mode":"uaa_auth","self_registration":"false","uaa_client_id":"harbor-cluster1","uaa_client_secret":"XXXXX","uaa_endpoint":"https://pks.XXX.wtf:8443","uaa_verify_cert":"true"}'

```

values.yaml
```diff
@@ -94,6 +94,24 @@ expose:
 # If Harbor is deployed behind the proxy, set it as the URL of proxy
 externalURL: https://core.harbor.domain
 
+## set if you're using external services with a self signed cert
+# authCACert: |
+#   -----BEGIN CERTIFICATE-----
```
ywk253100: We should keep the same style for all the certificates used in Harbor such as https://github.com/goharbor/harbor-helm/blob/master/values.yaml#L18. Maybe we can rename the `authCACert` to `uaaCACert` and refine the comment?

ywk253100:
@paulczar Thank you for the PR. I have added some comments for the UAA CA certificate part. 
For the configuration API, I think we can cover it in the document, such as `README`, rather than including it in the chart.

values.yaml
paulczar: Yes that makes sense to be able to look for it in an existing secret, I'll look to update that.

the reason I named it `authCACert` is that I assumed that at some point we might also support SSL for ldap/oauth/oidc and we could use this variable to provide the ca-cert for any of them, rather than needing a separate variable name each.

paulczar:
thanks @ywk253100 I feel quite strongly that the helm chart should be able to configure the whole application and there shouldn't need to be any extra steps, especially having to figure out how to `curl -x POST` a bunch of stuff into a configuration API.   

This PR is a compromise that allows a person to easily set the user config settings in `values.yaml` and have the notes section print out the appropriate `curl` command to run.  Perhaps in the future this could be done as a post-install job and remove the need for the user to run anything extra.

paulczar:
@ywk253100 I've updated the PR to rename the UAA CA Cert and optionally allow the user to set a secret name ( and error out of they try and supply both).  I also removed the user config settings, which I'll add as a separate PR.

values.yaml
ywk253100: Can we just keep the `uaaSecretName` and remove the `uaaCACert` as other secrets cannot be set as the content of the certificate?

ywk253100: And please squash your commits

paulczar:
ugh squashing it caused github to think I've included a bunch of other commits from previous people.  will close and recreate.

PR #254, opened 2019-06-09
Add userSettings section for helping deployer use config API

paulczar:
This makes it a little bit easier for the person configuring Harbor as it will provide a place to put user settings to be displayed in NOTES.txt for hints on using the config api via curl.


tcarmet:
Hello @paulczar and @ywk253100 

I already have a working patch that creates a Kubernetes `Job` with the same command.

I can happily contribute it on top of your changes if you would like to go forward on that direction.

Just let me know :)

ywk253100:
I like this idea, thanks @tcarmet.
@paulczar I think this is a better option, what do you think?

paulczar:
I think running `curl` commands at the API could be a bit fragile to automate.  I'd rather see work on a harbor-cli that could be run to set those configs, I've seen a proposal for a harbor-cli in the community repo.

I'd rather just be able to set this stuff via a config file or environment variable as that's the way most software is configured, and most automation tooling handles this sort of thing.  Introducing this configuration API without a decent client to use is making it much harder to automate.

Hokwang:
My opinion is that both of yours are not easy.
I just want to 
simply user can add/edit some vaules in values.yaml
and it works directly, not curl or not harbor-cli, etc.

like codes in core-cm.yaml
https://github.com/goharbor/harbor-helm/blob/master/templates/core/core-cm.yaml#L47-L51

paulczar:
@Hokwang I agree!  See this issue - https://github.com/goharbor/harbor/issues/8076

templates/NOTES.txt
```diff
@@ -1,3 +1,14 @@
...
-For more details, please visit https://github.com/goharbor/harbor.
\ No newline at end of file
+For more details, please visit https://github.com/goharbor/harbor.
+
+{{ if .Values.userSettings }}
+If this is your first time Installing Harbor,
+You may need to use the Harbor Configuration API to
+finish your configuration by running:
+
+$ curl -i -X PUT -u "admin:{{ .Values.harborAdminPassword }}" \
```
hobti01: I like the idea of the hint, but revealing the admin password in output that could be going to a pipeline deployment is not a secure idea.

templates/NOTES.txt
```diff
@@ -1,3 +1,14 @@
...
+  -H "Content-Type: application/json" \
+  {{ .Values.externalURL }}/api/configurations \
+  -d '{{ .Values.userSettings | toJson }}'
```
hobti01: Same concern here, dumping the client secrets and other sensitive data to deployment output is a risk.

PR #256, opened 2019-06-11
Add a config job that setup harbor's main configuration

tcarmet:
This pr relates to https://github.com/goharbor/harbor-helm/pull/254

Setting up via helm a config job that post Harbor's settings
into the configuration API via the parameters set on the `values.yaml`

Signed-off-by: Thomas Carmet <thomas.carmet@scality.com>


values.yaml
```diff
@@ -474,3 +474,12 @@ redis:
...
   ## Additional deployment annotations
   podAnnotations: {}
+
+jobConfig:
+  # if enabled setup proper Harbor settings
+  # https://github.com/goharbor/harbor/blob/master/docs/configure_user_settings.md#harbor-user-settings
+  enabled: false
+  settings: {}
+  image:
+    repository: byrnedo/alpine-curl
```
tcarmet: Here unfortunately I wasn't able to find an "official" curl docker image.

Tell us how to move forward with an official image hosted by goharbor per example.

paulczar:
I feel like wrapping a bunch of `curl -X POST` inside a job to do config automation might be a bit fragile, ideally we'd want to deal with errors and issues if the curl fails.  I think for now providing the command for the operator to run in the NOTES.TXT section is a starting point.

tcarmet:
Hello @paulczar,

Thank you very much for your response,

What we were trying to achieve here is to fully automate our installation of Harbor. The only thing that was missing was to setup our SSO service. Following the principle of Infrastructure as code, we consider that SSO settings should come with the install, and not require an additional manual step.

Now getting back to the subject of fragility: 
* Kubernetes gathers all the logs, so we can investigate if there's a failure.
* Thanks to Kubernetes retry mechanism, if the curl fails because the API is not available yet (or any other reason), it will retry.
* As the job is deactivated by default, the operator has to configure it, forcing the awareness of this job mechanism.

Is there any specific consideration that we might be missing, which would explain the fragility of the job based mechanism?

Now I do agree that having a `harbor-cli` as you were saying in another PR, would be great to do this kind of job, because if the way you interact with the API changes, we could apply the changes with it. but as it's not available yet, having this simple curl can be really useful.

Thanks! :)

paulczar:
I agree with what you're trying to achieve @tcarmet !  I just wish we didn't have to.  I wouldn't fight this PR going in, I just dislike the fact its even needed.

pre 1.8 it was dead simple as you could set the config settings all via ENV vars or config file ... all automation tooling know how to do both of those things.  whether its chef or puppet,  or something like confd pulling from consul. Even better languages like spring have integration to use kubeconfig maps as config sources.

These are all things we've learned to treat as code, and as devops/sre/whatever we have built up strong muscles around managing, testing, securing, etc, treating config changes as "code changes".

With the new configuration API however we cannot do any of those things, and instead of tried and true trusted ways to configure our app, the developers are forcing us to use an API without a decent client tool.  now instead of the simple task of "write file -> restart service" we have to rewrite automation tooling to take our list of preferred settings and sling them into the API via curl or similar.

tcarmet:
Thanks again for sharing your thoughts @paulczar 

I didn't knew we could apply those changes via config files and env variables. Indeed we wouldn't be having this conversation on pre 1.8 :smile: 

I 100% agree with you, we learned to treat those config as code, and it's hard to tear that apart.
On the other side, I like having an API for those config, as it doesn't require any service restart, which is pretty cool.

I've tried thinking about how we could do this in a "devops" friendly way, and I stubbled upon [this issue](https://github.com/goharbor/harbor/issues/8020), which talks about making an operator for Harbor. Maybe, this is the kind of tasks an Kubernetes operator could do, handling config like this one, via CRDs.

paulczar:
see this issue - https://github.com/goharbor/harbor/issues/8076

PR #266, opened 2019-06-19
Adds in RBAC, SecurityContext, and PSP groundwork

voor:
This adds in a unique service account for each deployment or statefulset in Harbor, and the accompanying PodSecurityPolicies.

RBAC and PSP are at the moment global, and applied from the values entry, although depending on the permissions for each part they might need to be broken out.

Currently this is failing, since every container is running as root and still tries to `sudo` at startup.


voor:
This is more or less functioning.  Unfortunately, the underlying containers are doing a lot of file changing ownerships and `sudo` calls that are going to make locking down some of these security contexts a lot more involved.  For now, this at least creates some of the missing pieces to allow installation into a cluster that explicitly removes default PSP.

voor:
Specifically referenced removing the sudo stuff in this issue: https://github.com/goharbor/harbor/issues/8300

uLan08:
Would love to see this merged. If there's anything you need help, let me know!

bradkwadsworth:
This would be wonderful if the ability to add annotations to the ServiceAccount objects were also added. It would make things like enabling IAM Roles for Service Accounts in AWS possible.

mkuendig:
How do we go with this? I would need this too.

HugoRh:
I am also in a case where it would be very convenient to have it!

jonkerj:
Would love to see this merged. What would it take to have someone from Harbor to look at this PR? It has been open for quite a while now

jonkerj:
Hello? Anybody out there?

aemaem:
Does anyone know the status here? This would be very much appreciated!

PR #271, opened 2019-06-27
add registry.relativeurls value

### LINKED ISSUE #174: "unknown blob" error while pushing images to harbor
I have successfully deployed harbor-helm on my openshift cluster. I am able to login to harbor. However, when i tried to push image to harbor repository. The "unknown blob" appears after completing the command. 
Here is the output of my command:
[root@master harbor-helm]# docker push harbor.192.168.80.130.nip.io/library/alpine:v1.0
The push refers to a repository [harbor.192.168.80.130.nip.io/library/alpine]
503e53e365f3: Pushing [==================================================>] 5.529 MB/5.529 MB
unknown blob
Kindly help me in this regard.

pablo-ruth:
Add a value to config relativeurls support on registry. This is needed when expose.type is not set to ingress to avoid "unkown blob" error on image push.

Resolves #174

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -374,6 +374,8 @@ registry:
   # If a secret key is not specified, Helm will generate one.
   # Must be a string of 16 chars.
   secret: ""
+  # If true, the registry returns relative URLs in Location headers. The client is responsible for resolving the correct URL.
+  relativeurls: false
   middleware:
     enabled: false
     type: cloudFront
--- a/templates/registry/registry-cm.yaml
+++ b/templates/registry/registry-cm.yaml
@@ -143,6 +143,7 @@ data:
       addr: "{{ template "harbor.redis.host" . }}:{{ template "harbor.redis.port" . }}"
       db: {{ template "harbor.redis.registryDatabaseIndex" . }}
     http:
+      relativeurls: {{ .Values.registry.relativeurls }}
       addr: :5000
       # set via environment variable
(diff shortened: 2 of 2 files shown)


BartJoris:
Hi, any updates on this?

deschmih:
I need this too, because same error occurs when running the registry on non default port. So this fix would help.

ywk253100:
@pablo-ruth Could you resolve the conflict?

values.yaml
```diff
@@ -353,6 +353,8 @@ registry:
   # If a secret key is not specified, Helm will generate one.
   # Must be a string of 16 chars.
   secret: ""
+  # This needs to be true when expose.type is not ingress
```
ywk253100: I don't think it's related with the expose.type. Could you update the comment according to the document of docker registry https://github.com/docker/distribution/blob/master/docs/configuration.md#http?

pablo-ruth:
Ok done :)

ywk253100:
@pablo-ruth Could you squash all commits into a single one?

pablo-ruth:
Done, commits are squashed

ywk253100:
@pablo-ruth Thanks!

problame:
ISTM that this needs to be back-ported to the component that generates `docker-compose.yml`. 

ref https://github.com/goharbor/harbor-helm/issues/174#issuecomment-484161686

PR #273, opened 2019-06-28
Add sideCar container support

### LINKED ISSUE #247: [Contribution proposal] Providing a SideCar container utility for External DB (GKE?)
Hello,

For my company, I'm trying to setup Harbor using your chart (v1.1.0), deployed in a GKE, using a Google SQL (postgre) instance as the Database.

By default, Google Cloud suggests to use the Public DB IP, which makes no sense when the SQL Instance as well as the Harbor services are internal to GKE.
Alternatives include:
 - Using private IP for the SQL Instance (Beta version for our Infra-as-code tools)
 - Using a sidecar Container (we're already using this method for other services).

We thought that using a sidecar container to proxify the access to the database would be an appropriate solution, and already have a patch for this internally.

Our internal patch is currently only handling external database, on the premise that it should be configured in the "database" block of the values.yml file from the chart.

It could be changed if you (as a project handling team) have a preferred alternative.
Would you be interested in such a contribution ?

ghost:
This sideCar container support is done through the use of a set of two
variables, applied individually for each pod in addition to a global
processing.

These two variables are the relatively de-facto standard `extraContainers` and
`extraVolumes` (for the extraContainers)

Fixes #247

### DIFF
--- a/templates/chartmuseum/chartmuseum-dpl.yaml
+++ b/templates/chartmuseum/chartmuseum-dpl.yaml
@@ -76,6 +76,12 @@ spec:
           mountPath: /harbor_cust_cert/custom-ca-bundle.crt
           subPath: ca.crt
         {{- end }}
+      {{- if .Values.extraContainers }}
+{{ toYaml .Values.extraContainers | indent 6 }}
+      {{- end }}
+      {{- if .Values.chartmuseum.extraContainers }}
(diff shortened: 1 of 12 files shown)


DavidPineauScality:
updated.

DavidPineauScality:
Updated

PR #274, opened 2019-06-28
Fix GC 500 error when using with imageChartStorage 'gcs'

alexnguyen91:
Garbage collector got 500 error when using GCS to store image. The reason is because missing mount for gcs-key.json so this PR is an attempt to fix it.

### DIFF
--- a/templates/registry/registry-dpl.yaml
+++ b/templates/registry/registry-dpl.yaml
@@ -134,6 +134,11 @@ spec:
           mountPath: /harbor_cust_cert/custom-ca-bundle.crt
           subPath: ca.crt
         {{- end }}
+        {{- if and .Values.persistence.enabled (eq .Values.persistence.imageChartStorage.type "gcs") }}
+        - name: gcs-key
+          mountPath: /etc/registry/gcs-key.json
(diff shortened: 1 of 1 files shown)


EvertonSA:
@alexnguyen91 can you help me out? The docs says: 
> # The base64 encoded json file which contains the key
>     encodedkey: base64-encoded-json-key-file

What is this key? I tried the SA GCS key, the GCS HMAC key, encode and not encoded but everytime is returns me:
> release harbor failed: Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 0, error found in #10 byte of ...|dmin.json","REGISTRY|..., bigger context ...|"data":{"GCS_KEY_DAT

ywk253100:
@EvertonSA The error said it isn't a valid base64 encoded data

ywk253100:
@alexnguyen91 Sorry for the late reply. Could you fix the conflict?

alexnguyen91:
@ywk253100 Could you check please ?

ywk253100:
@alexnguyen91 Thanks!

PR #281, opened 2019-07-04
Additional options with using expose type of LoadBalancer

jamessthompson:
Added support for annotations and source IP address ranges against LoadBalancer type service

Signed-off-by: James Thompson <jst2120@gmail.com>


ywk253100:
@jamessthompson Thanks.

PR #284, opened 2019-07-07
updates docs with correct keys to use for notary tls Secret

chrisfarms:
## What

Update the README to reflect the correct values required when generating a notary tls secret.

## Why

when using a custom notary.secretName secret to avoid re-generating notary secrets on each template render; the docs said to populate the secret with "tls.ca", "tls.crt" and "tls.key" key names... this was incorrect as the template appears to expect "ca", "crt" and "key" key names.


ywk253100:
@chrisfarms Thanks for the PR. 
Could you change the key names in templates to `tls.crt`, `tls.key` and `ca.crt` to keep consistence with other certifate?

BTW, the DCO is required for the PR

ywk253100:
@chrisfarms ping

ywk253100:
Fixed by #244, closing this one

PR #291, opened 2019-07-17
Allow to overwrite registry auth

robinro:
This makes it possible to use other authentication services or another ingress for the core service to access the registry.


robinro:
We run the harbor ui behind an oauth proxy. With this change we can configure an alternative auth that allows to authenticate to the registry inside the same cluster via the internal service name of the harbor core service instead of via the externalURL.

ywk253100:
This doesn't apply to Harbor v2.x anymore, closing this PR.

PR #295, opened 2019-07-19
Support AWS ALB ingress controller

vshimoda:
AWS ingress controller must be NodePort service. And / must be the last because the path matching depends on the order.


templates/ingress/ingress.yaml
```diff
@@ -89,6 +91,12 @@ spec:
         backend:
           serviceName: {{ template "harbor.core" . }}
           servicePort: 80
+      {{- if eq .Values.expose.ingress.controller "alb" }}
```
ywk253100: What is it used for?

values.yaml
```diff
@@ -31,6 +31,7 @@ expose:
     # leave as `default` for most ingress controllers.
     # set to `gce` if using the GCE ingress controller
     # set to `ncp` if using the NCP (NSX-T Container Plugin) ingress controller
+    # set to `alb` if using the AWS ALB ingress controller
```
ywk253100: Please add it to `README` as well

templates/ingress/ingress.yaml
vshimoda: In ALB ingress controller, path matching rules are applied in order from the top.  "/" seems to be recognized as "/ *", so rules after "/" will not work.

horjulf:
Need to expose service annotations too, so that the health check path for the core service can be set correctly (defaults to / which gets a 404) or set the annotation by default.

`alb.ingress.kubernetes.io/healthcheck-path: /api/ping`

What do you think @ywk253100 ?

templates/ingress/ingress.yaml
ywk253100: How about put the portal section as the last one? this should work for other ingress controllers

ywk253100:
@horjulf Didn't get your point. Why we need to check the health of service?

ywk253100:
@vshimoda Could you fix the conflict and add the DCO?

horjulf:
> @horjulf Didn't get your point. Why we need to check the health of service?

By default the target groups created by the ALB ingress controller in AWS, have default health checks (default path `/`).
For example the core service responds with a 404 on `/`, making the target group healthcheck fail.

image

Lemmons:
It appears this work has stalled. Any chance of it getting merged?  This is definitely needed.

zyyw:
@vshimoda , thanks for contributing to Harbor-helm. Could you please rebase latest master into your branch, and then update your PR accordingly? Thanks

jamesglennan:
Also very interested in this, currently have manually edited the chart, but it would be cool have this supported natively

templates/ingress/ingress.yaml
```diff
@@ -1,7 +1,7 @@
 {{- if eq .Values.expose.type "ingress" }}
 {{- $ingress := .Values.expose.ingress -}}
 {{- $tls := .Values.expose.tls -}}
-{{- if eq .Values.expose.ingress.controller "gce" }}
+{{- if or (eq .Values.expose.ingress.controller "gce") (eq .Values.expose.ingress.controller "alb") }}
```
jamesglennan, line 4: I just tested this and it doesnt work (maybe changes since this was first worked on)
It seems alb controller doesnt like wildcard

I was able to get ingress going with changes to the service (svc) templates so that when alb is specified it makes them nodeport, but the ingress rules stay the same

jamesglennan:
This PR can be closed with the merge of #1206

zyyw:
close this PR because of the merge of the following PR:
* https://github.com/goharbor/harbor-helm/pull/1206

PR #298, opened 2019-07-24
Add possibility for custom pod labels

KarstenSiemer:
Adds the possibility to add custom pod labels for people who like to use the kubernetes recommended labels like here:
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Reason being NetworkPolicies that could be set as a default inside a namespace that wants to attach to some label. Users should have the possibility to set such label in the values directly

### DIFF
--- a/templates/chartmuseum/chartmuseum-dpl.yaml
+++ b/templates/chartmuseum/chartmuseum-dpl.yaml
@@ -16,6 +16,9 @@ spec:
     metadata:
       labels:
 {{ include "harbor.labels" . | indent 8 }}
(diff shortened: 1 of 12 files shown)


Hokwang:
I use redis externally, and there is comment
```
Note: Since NetworkPolicy is enabled, only pods with label
harbor-redis-client=true"
will be able to connect to redis.
```

so we need to use label.

@ywk253100

sidewinder12s:
Could this get rebased? We've also got similar requirements for labeling pods to track cost information.

PR #318, opened 2019-08-14
Expose health endpoint of each component.

MartinForReal:
add health check point for nginx, distribution and clair 
https://github.com/goharbor/harbor/issues/4557


MartinForReal:
already included in new config

PR #319, opened 2019-08-14
update nodePort usage with Helm

kacole2:
Simple changes to update nodePort docuementation

Signed-off-by: Kendrick Coleman <kendrickc@vmware.com>


values.yaml
```diff
@@ -94,7 +94,7 @@ expose:
 # 2) if "expose.type" is "clusterIP", the "domain" should be
 # the value of "expose.clusterIP.name"
 # 3) if "expose.type" is "nodePort", the "domain" should be
-# the IP address of k8s node
+# the IP address or DNS of k8s node with nodePort (ex https://IPorDNS:30003)
```
ywk253100, line 97: The `domain` should be IP and DNS and the nodePort should be specified in the `port` part

PR #324, opened 2019-08-19
Run as non-root user

ywk253100:
Add init containers to change permission of directory to allow the progress runs as non-root user

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -28,6 +28,34 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
+      - name: "change-permission-of-directory"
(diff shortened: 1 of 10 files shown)


templates/core/core-dpl.yaml
```diff
@@ -28,6 +28,34 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
+      - name: "change-permission-of-directory"
+        image: {{ template "harbor.changePermissionImage" . }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        command: ["/bin/sh"]
+        args: ["-c", "chown -R 10000:10000 /etc/core || true"]
+        volumeMounts:
```
reasonerjt: Is there any pattern in k8s to avoid such duplication?  Similar to `volume-from`

PR #332, opened 2019-08-21
Update values.yaml to add new attributes introduced in 1.9

ywk253100:
Update values.yaml to add new attributes introduced in 1.9


ywk253100:
Fixes https://github.com/goharbor/harbor/issues/8409

PR #339, opened 2019-09-05
add support for AWS S3 Server Side Encryption

usualstuff:
This commit adds support for AWS S3 SSE for registry and chartmuseum. It's already defined in values file (see "keyid").


torstenwalter:
Looks like a great addition. I know customers who would require encryption at rest.

unguiculus:
Any chance we can get this in?

PR #355, opened 2019-09-18
expose health endpoint of each component.

MartinForReal:
Signed-off-by: Fan Shang Xiang <fanshangxiang@mskj.com>


ywk253100:
The Harbor chart won't provide any capabilities that isn't included in the OSS harbor.
If this is valuable for you, yon can try to open an issue in the harbor repository https://github.com/goharbor/harbor and discuss this with the maintainers.

Issue #224, opened 2019-05-02
Harbor helm upgrade fails due to volume multi attach error on multiple workers


yakbulut:
Hi all,

We are running into an issue where upgrading harbor chart from 1.0.0 to 1.0.0 causes upgrade to break as JobService and Registry pods are stuck in ContainerCreate. Looking into kubectl describe reveals these have failed due to not being able to mount the volume.`Multi-Attach error for volume "pvc-xxx" Volume is already used by pod(s)`

Essentially, what happens is that during an upgrade if these pods are scheduled to start on different workers, new pods are unable to mount the volumes due to RWO access. We are unable to change this to RWM as it is not supported by vSphere cloud provider.

I have found the workaround to be manually setting the replica to 0 to terminate registry and jobservice and then patch it back to 1 right after upgrade. However, this is less than ideal.

Repro steps:
1. Get a kubernetes cluster with multiple workers
2. Helm install harbor 1.0.0
3. Helm upgrade harbor using the same values with --force and --recreate-pods
4. Registry and JobService are stuck at ContainerCreate

What is the solution for this issue ? Does harbor chart only work with RWM access on in a multi worker setup ?

Thanks,

Benjvfr:
Hi,

Same problem again concerning the upgrade from Helm Chart v1.2.1 to v1.2.2.

Do you know how to that ?

Thanks.

ywk253100:
Fixed

PR #366, opened 2019-09-20
If PVC is used, force only one replica in deployments

YorikSar:
This enables proper upgrade when PVC is bound to existing replica.

Fixes: #224

### DIFF
--- a/templates/chartmuseum/chartmuseum-dpl.yaml
+++ b/templates/chartmuseum/chartmuseum-dpl.yaml
@@ -7,7 +7,13 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
     component: chartmuseum
 spec:
+  {{- if and .Values.persistence.enabled (eq .Values.persistence.imageChartStorage.type "filesystem") }}
+  replicas: 1
+  strategy:
+    type: Recreate
+  {{- else }}
   replicas: {{ .Values.chartmuseum.replicas }}
+  {{- end }}
   selector:
     matchLabels:
 {{ include "harbor.matchLabels" . | indent 6 }}
--- a/templates/jobservice/jobservice-dpl.yaml
+++ b/templates/jobservice/jobservice-dpl.yaml
@@ -6,7 +6,13 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
     component: jobservice
 spec:
+  {{- if and .Values.persistence.enabled (eq .Values.jobservice.jobLogger "file") }}
+  replicas: 1
+  strategy:
+    type: Recreate
+  {{- else }}
   replicas: {{ .Values.jobservice.replicas }}
+  {{- end }}
   selector:
     matchLabels:
 {{ include "harbor.matchLabels" . | indent 6 }}
--- a/templates/registry/registry-dpl.yaml
+++ b/templates/registry/registry-dpl.yaml
@@ -6,7 +6,13 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
     component: registry
 spec:
+  {{- if and .Values.persistence.enabled (eq .Values.persistence.imageChartStorage.type "filesystem") }}
+  replicas: 1
+  strategy:
+    type: Recreate
+  {{- else }}
   replicas: {{ .Values.registry.replicas }}
+  {{- end }}
   selector:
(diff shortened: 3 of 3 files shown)


mumrau:
I like the strategy, I chose the other path which is to switch these workloads to `StatefulSet` which carries the meaning of something... well stateful, because it depends on the state of your actual pvc.

Looking forwrad from the maintainers, as it became quite bothering to fork the repository at ever release.

templates/chartmuseum/chartmuseum-dpl.yaml
```diff
@@ -7,7 +7,13 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
     component: chartmuseum
 spec:
+  {{- if and .Values.persistence.enabled (eq .Values.persistence.imageChartStorage.type "filesystem") }}
+  replicas: 1
```
ywk253100: Why the replicas must be 1 here?

templates/chartmuseum/chartmuseum-dpl.yaml
```diff
@@ -7,7 +7,13 @@ metadata:
...
+  strategy:
```
ywk253100: How about we expose the `strategy` as a configuration item in `values.yaml`?

ywk253100:
@YorikSar Sorry for the late reply. Please see my comments

templates/chartmuseum/chartmuseum-dpl.yaml
YorikSar: If `imageChartStorage.type` is `filesystem` then storage is done as regular files, and you can't share one ("dumb") filesystem across multiple replicas, so we have to make sure that no more than one replica is created.

templates/chartmuseum/chartmuseum-dpl.yaml
YorikSar: Since we can't have more than one replica up at a time, we have to destroy old replica before creating a new one, which is what `Recreate` strategy does. We could expose strategy details for HA (multi-replica) case in a separate change.

templates/chartmuseum/chartmuseum-dpl.yaml
ywk253100: The `imageChartStorage.type` is an application level configuration item, it can be `filesystem` or other cloud storage. But even though it's `filesystem`, it's backend PV can be shared between multiple nodes. So the `replicas: 1` isn't needed

templates/chartmuseum/chartmuseum-dpl.yaml
ywk253100: Thanks. I think we can still use the `RollingUpdate` as the default value. 
And please add some comment for the new added properties in both `value.yaml` and `README` to explain what it 's used for

ywk253100:
@YorikSar Hi YorikSar, we are about to release a new version of Harbor chart in next week, it's good if this fix can be included. Do you have any time to update the PR according to the discussion? 

Please let me know if you are busy with other things, I'll update it based on your changes.

templates/chartmuseum/chartmuseum-dpl.yaml
YorikSar: Unfortunately this would require PVC with ReadWriteMany which is working only on small subset of volume plugins.

templates/chartmuseum/chartmuseum-dpl.yaml
YorikSar: With RollingUpdate this Deployment will never get updated unless PVC has ReadWriteMany mode:
- first new pod is created
- new pod waits for PVC to get released by old pod
- old pod is not being killed by Deployment, because new one is not up yet.

YorikSar:
@ywk253100 Unfortunately I don't have time to adjust and test this PR right now, feel free to take it over.

templates/chartmuseum/chartmuseum-dpl.yaml
ywk253100: Yes. So setting replicas to 1 or not is decided by users

ywk253100:
PR https://github.com/goharbor/harbor-helm/pull/470 is submitted, closing this one, thanks again @YorikSar

PR #367, opened 2019-09-23
ingress: "nosecret" option

multi-io:
This adds an option to leave out the "secretName" setting in the tls section of generated ingress resources. This is useful for setups where the ingress controller already provides a default TLS secret for multiple or all ingresses in the cluster, which is supported e.g. by the nginx-ingress-controller via its `--default-ssl-certificate` option.


tmojlupo:
I was just about to submit an issue, but found this PR that would solve our problem.  Our ingress configuration provides a default certificate for all ingress resources unless otherwise specified.  So, we currently have to do some manual edits to strip out the ingress cert and secret that this Helm chart creates.

Can this get some attention?

PR #370, opened 2019-09-24
set proxy_request_buffering off

liuxu623:
Fixes https://github.com/goharbor/harbor/issues/9217


mac-chaffee:
I think this is quite an important change to make, thanks for submitting this @liuxu623 !

Another person also encountered the same issue and explained it: https://github.com/distribution/distribution/issues/2634

Essentially, we really do need to disable request buffering because pushing many large image layers at once can kill an ingress-nginx instance due to buffering too much data to disk. If ingress-nginx isn't run with ephemeral-storage limits (the default), then too much buffering can even destabilize a whole Kubernetes node due to disk pressure which triggers cascading evictions.

Seems the maintainers haven't noticed this PR; want me to recreate it for you @liuxu623 ?

PR #372, opened 2019-09-24
Add option to enable metrics for harbor registry

mvandergiesen:
Make it possible to enable metrics on the registry service and expose the debug port.


ywk253100:
Dup of https://github.com/goharbor/harbor-helm/pull/205

Issue #371, opened 2019-09-24
Harbor upgrade v1.8 to v1.9 issues


liquid-cloud:
We've upgraded our registry from v1.8.1 to v.1.9.0.

During the upgrade phase, we encountered the following:

- The pods were getting terminated before the migration was complete. Reason being that the threshold for the liveliness probe was too small to allow the migration to complete. The default `failureThreshold` is `3` ([reference](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes))  which makes the core container to get killed after one minute if it's not ready according to [the code](https://github.com/goharbor/harbor-helm/blob/master/templates/core/core-dpl.yaml#L37-L42).
Example log from the core container when this happens:
```
2019-09-24T07:11:13Z [INFO] [/core/main.go:146]: capture system signal terminated, to close "closing" channel
2019-09-24T07:11:16Z [INFO] [/core/main.go:152]: Timeout waiting goroutines to exit
```

- Having multiple core containers with version 1.9 starting up (possibly from scaling down and scaling up the deployment manually) can lead to database corruption. We've learned this the hard way unfortunately. We managed to restore a snapshot of the database and continue with the upgrade.

During the upgrade I suggest you take into consideration on how slow the storage is. We have around 400GB image data and 5000+ artifacts in total.

It took roughly ~7 minutes to persist the data for a project with ~1800 artifact count and ~210GB quota according to the new UI.

In total it the upgrade took ~20 minutes to persist the data:
```
2019-09-20T09:41:30Z [INFO] [/core/api/quota/migrator.go:111]: [Quota-Sync]:: start to persist data for server ... [registry]
.......
2019-09-20T09:59:07Z [INFO] [/core/api/quota/migrator.go:116]: [Quota-Sync]:: success to persist data for server ... [registry]
2019-09-20T09:59:10Z [INFO] [/core/main.go:107]: Success to sync quota data .....
2019/09/20 09:59:10 [I] [asm_amd64.s:1337] http server Running on http://:8080
```

yevhenvolchenko:
I think, that liveliness probe implemented incorrectly.
Now it's pointing at the same endpoint, as readiness does, so when the pod is running migrations is in non-live state, when it actually does.
**Pod should return 200 on liveliness during migrations** to avoid this issue in future.

ywk253100:
@yevhenvolchenko Good point!
[...]

PR #380, opened 2019-10-04
Make core probes configurable

bpehling:
Make core probes configurable. On quota migration the core pods require more time with large amounts of data and should be configurable to allow migration to complete.

Issue: On quota migration from 1.7 to 1.9 the pods were restarting since the checks weren't ready and restarted the containers which resulted in restarting the migration before completion.

Signed-off-by: Ben Pehling <shinjipehling@gmail.com>


ywk253100:
The PR is fixing #371, we'll estimate the solution provided by the PR. Thanks @bpehling

values.yaml
```diff
@@ -300,6 +300,12 @@ core:
     repository: goharbor/harbor-core
     tag: dev
   replicas: 1
+  livenessProbe:
```
ywk253100: @bpehling I think we can just make the `initialDelaySeconds` of `livenessProbe` configurable to fix the upgrading issue and with less configuration items for users.

Please add comment for the item to explain what is it for and document it in the `README.md`

ywk253100: And larger the delay time, maybe 300 is a reasonable value?

bpehling: I think 300 is reasonable, fixed

ywk253100: Thanks

ywk253100:
@bpehling Could you cherry pick the commit into the `1.2.0` branch? We'll release 1.2.1 to include some fixes.

PR #407, opened 2019-10-21
Update Harbor chart to enable pluggable scanner

ywk253100:
Update Harbor chart to enable pluggable scanner


reasonerjt:
Now that clair-adapter and clair is in one pod, the service should only expose clair-adapter's API?

templates/core/core-cm.yaml
```diff
@@ -36,6 +36,7 @@ data:
   CLAIR_DB: "{{ template "harbor.database.clairDatabase" . }}"
   CLAIR_DB_SSLMODE: "{{ template "harbor.database.sslmode" . }}"
   CLAIR_URL: "http://{{ template "harbor.fullname" . }}-clair:6060"
+  CLAIR_ADAPTER_URL: "http://{{ template "harbor.clair" . }}:8080"
```
reasonerjt: We should only need clair_adapter_url in core's CM?

templates/clair/clair-dpl.yaml
```diff
@@ -53,16 +53,41 @@ spec:
...
+          periodSeconds: 10
+        readinessProbe:
+          httpGet:
+            path: /probe/ready
+            port: 8080
+          initialDelaySeconds: 30
+          periodSeconds: 10
+        env:
+        - name: SCANNER_CLAIR_URL
+          # To avoid a pod cannot reach itself via service IP when the clusters disable hairpin
```
reasonerjt: I think it's totally fine as they are always in the same pod and share the same network.

templates/core/core-cm.yaml
ywk253100: We'll remove this variables after they are deleted from main repository

Issue #260, opened 2019-06-17
StatefulSet options for Deployments with PVC (chartmuseum for instance)


mumrau:
Hello,

As stated in #7 , a few components moved to Deployment from StatefulSet because they were stateless, at least at the time.
Unfortunately, a few things changed and some of the components now can be persistent, (I say can, because for example chartmuseum can be used with various cloud storage backends) and they use PVC which are most likely in ReadWriteOnce.
Thus, when you apply a change, another ReplicaSet gets spawned, and since it cannot mount the volume cause it is already mounted on the old ReplicaSet, we're stucked. The old one isn't deleted because the new one cannot reach a ready state.
My possible fix would simply be to give the choice in between SS and Dpl for some components.

Could somebody brigthen me about this?

ywk253100:
@mumrau As we discussed in https://github.com/goharbor/harbor-helm/pull/410, the `jobservice`, `chartmuseum` and `registry` are both Deployment rather than StatefulSet.
And a new PR https://github.com/goharbor/harbor-helm/pull/470 is submitted to fix the upgrade issue, closing this one.

PR #410, opened 2019-10-22
Single PVC bound workload capacity can now be statefulSet on demand

mumrau:
Hello,

This PR gives the end-user the possibility to chose whether their workload for jobService, registry, and chartmuseum are `StatefulSets` or `Deployments`.

This is extremely handy and necessary in case we use a `ReadWriteOnce` `PersistentVolume`, indeed, on rolling upgrades, the pvc stays bounded to the pod of the "old replicaset", and thus the pod from the "new replicaset" cannot bind the volume. We need to manually delete the old replicasets for these 3 workloads, which makes it very hard to maintain multiple consistent harbor installations accross clusters.

It solves #260 , https://github.com/goharbor/harbor/issues/8181, and is a different approach than #366 which makes more sense to me. Indeed, if you use RWO volumes, the workload is _stateful_ by nature, and thus deservers the type `StatefulSet`.

I am using this little tweak since harbor 1.7 and have successfully updated all the way to 1.9.1.


tlvenn:
Hi @ywk253100 , could you please review this PR and merge it if you find it acceptable ? It would be great to include it in the next public 1.2 chart.

Thanks in advance.

tlvenn:
Hi @mumrau,

Thanks for this PR, I think this is the right solution however when a statefulSet is used, we should really be using a volumeClaimTemplate and not create a PVC like we do for a deployment.

Could you update your PR to account for this ? This would be great !

ywk253100:
@tlvenn Sorry for the late reply. We decide to use the fix in PR https://github.com/goharbor/harbor-helm/pull/366. We'll expose the upgrade strategy configuration item.

tlvenn:
No worry, do you have a target date for the 1.3.0 release ?

ywk253100:
It should be on early of Dec if everything goes well

mumrau:
Indeed @tlvenn, it is fixed this way in the chart I use on my clsuters. I forgot to include it in the PR as well obviously.
It seems like this is going to be abandonned anyways.

I still find everything kind of odd, not using `StatefulSet` for something, well, `Stateful`.

tlvenn:
I think it might be interesting to propose it again on top of the merged PR instead of trying to compete with it.

StatefulSet are the way to go for all those scenarios (RWO pvc) imho but it is really hard to upgrade from a declared pvc to a volumeClaimTemplate given you don't control the created pvc name in the later. So when you upgrade an existing harbor installation, the other proposal while not being ideal does solve the problem seamlessly. For new harbor installation, going with StatefulSet is better and should be favored.

ywk253100:
@mumrau Not applications need persistent storage are all `StatefulSet`. When you scale the `StatefulSet`, the pods will use their own persistent volumes, not share a single one among all pods, this is not correct for `jobservice` and `registry`. 
The `jobservice` and `registry` are both `Deployment`  rather than `StatefulSet`.

tlvenn:
Ya that's a good point @ywk253100 and definitely thinking with a replicas > 1 helps understand which is which.

ywk253100:
PR https://github.com/goharbor/harbor-helm/pull/470 is submitted to fix the issue, closing this one

PR #412, opened 2019-10-22
Template database init container image for 'change-permission-of-directory'

maratoid:
* Allows for disconnected/offline installs

Cannot currently install harbor from an offline harbor repository without access to pull upstream container 'busybox:latest' (or pre-caching same on hosts)

* Reuses existing harbor database container by default

As init container is only executing chown, the existing database image should suffice for the init container, reducing the number of required images.

### DIFF
--- a/templates/database/database-ss.yaml
+++ b/templates/database/database-ss.yaml
@@ -31,7 +31,7 @@ spec:
       {{- end }}
       initContainers:
       - name: "change-permission-of-directory"
-        image: "busybox:latest"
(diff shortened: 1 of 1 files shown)


ywk253100:
@maratoid Thanks for your PR. But the database container runs with non-root user, it has no privilege to change the owner of directory. 
For the disconnected/offline installation, I think we can provide the configuration item for the image of init container in the `values.yaml`

ywk253100:
Fixed in PR https://github.com/goharbor/harbor-helm/pull/423, closing this one

Issue #416, opened 2019-10-29
Configure the log level for clair adapter


ywk253100:
Depends on PR https://github.com/goharbor/harbor/pull/9640

PR #420, opened 2019-10-30
Add log level to Clair adapter

ywk253100:
Add log level to Clair adapter


ywk253100:
Fixes #416

PR #422, opened 2019-10-30
Update installation part in HA document

ywk253100:
Use helm repo to download harbor chart


ywk253100:
Fixes #413

Issue #430, opened 2019-11-07
Add ignore media types for registry


ywk253100:
Depends on https://github.com/goharbor/harbor/pull/9756

PR #434, opened 2019-11-11
Add "ignoredmediatypes" to config.yml of registry

ywk253100:
The "ignoredmediatypes" ignores several notifications sent by registry


ywk253100:
Fixes #430.

PR #439, opened 2019-11-15
Bump up Harbor to 1.9.3

ywk253100:
Bump up Harbor to 1.9.3

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -268,7 +268,7 @@ proxy:
 nginx:
   image:
     repository: goharbor/nginx-photon
-    tag: v1.9.2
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -350,7 +350,7 @@ registry:
   registry:
     image:
       repository: goharbor/registry-photon
-      tag: v2.7.1-patch-2819-2553-v1.9.2
+      tag: v2.7.1-patch-2819-2553-v1.9.3
```
tlvenn: This image does not exist in Docker Hub

ywk253100: Just pushed

PR #463, opened 2019-12-06
Provide the configuration item xsrf key in values.yaml

### LINKED ISSUE #459: [bug] random string for XSRFKey causes core deployment update every time helm upgrade is run
There's a random string in the core configmap that causes the configmap to change every time `helm ugprade` is run.  This creates the undesirable effect of the core deployment always changing and thus  creating a new replicaset and new pods.   It also messes with anything using a `helm diff` in a gitops type workflow as it sees a change when there really shouldn't be one.

The offending commit is in Pull Request #431 

```
    EnableXSRF = true
    XSRFKey = {{ randAlphaNum 40 }}
    XSRFExpire = 3600
```

should be the following with a new value in the values.yaml `core.xsrfKey`.   you could possibly do something a bit cleaner, but this is verbose to show the working.

```
    EnableXSRF = true
    {{- if core.xsrfKey }}
    XSRFKey = {{ core.xsrfKey }}
    {{- else }}
    XSRFKey = {{ randAlphaNum 40 }}
    {{- end }}
    XSRFExpire = 3600
```

ywk253100:
Provide the configuration item xsrf key in values.yaml
Fixes #459


README.md
```diff
@@ -177,18 +177,19 @@ The following table lists the configurable parameters of the Harbor chart and th
...
+| `core.replicas` | The replica count  | `1` |
+| `core.livenessProbe.initialDelaySeconds` | The initial delay in seconds for the liveness probe | `300` |
+| `core.resources` | The [resources] to allocate for container | undefined |
+| `core.nodeSelector`  | Node labels for pod assignment | `{}` |
+| `core.tolerations` | Tolerations for pod assignment | `[]` |
+| `core.affinity` | Node/Pod affinities | `{}` |
+| `core.podAnnotations` | Annotations to add to the core pod | `{}` |
+| `core.secret` | Secret is used when core server communicates with other components. If a secret key is not specified, Helm will generate one. Must be a string of 16 chars. | |
+| `core.secretName` | Fill the name of a kubernetes secret if you want to use your own TLS certificate and private key for token encryption/decryption. The secret must contain keys named: `tls.crt` - the certificate and `tls.key` - the private key. The default key pair will be used if it isn't set | |
+| `core.xsrfKey` | The XSRF key, must be a string of 40 chars. Will be generated automatically if it isn't specified | |
```
reasonerjt: Does it allow special chars like `#` `$`?  I recall it does not?

ywk253100: Not find any clarification for that

Issue #464, opened 2019-12-09
The installation/delete commands in readme works only for helm v2.


reasonerjt:
We should add command example for managing the chart using helm v3

PR #466, opened 2019-12-10
Add commands for helm3 in doc

ywk253100:
Add commands for helm3 in doc


ywk253100:
Fixes #464

PR #469, opened 2019-12-12
add support for traefik ingressroute

rbtr:
This PR adds support for the [Traefik IngressRoute CRD](https://docs.traefik.io/v2.0/providers/kubernetes-crd/) as a method of ingressing traffic to Harbor.  

I was having trouble uploading layers to Harbor, getting an `unkown blob` 500 internal server error. Reading through the Harbor issues, that seems to show up when there's an HTTPS terminating proxy in front of the NGINX Harbor proxy, but even following the [troublshooting steps for this double proxy scenario](https://github.com/goharbor/harbor/blob/master/docs/installation_guide.md#using-nginx-or-load-balancing) I was not able to get Harbor to work behind Traefik. Since Traefik is doing what the NGINX is doing, but using the IngressRoute CRD instead of the Ingress object in this configuration, I thought it might be cleanest to enable direct Traefik CRD integrations.

### DIFF
--- a/templates/ingress/ingressroute.yaml
+++ b/templates/ingress/ingressroute.yaml
@@ -1,0 +1,47 @@
+{{- if eq .Values.expose.type "ingressRoute" }}
+{{- $ingressRoute := .Values.expose.ingressRoute -}}
+{{- $_ := set . "portal_path" "/" -}}
+{{- $_ := set . "api_path" "/api/" -}}
+{{- $_ := set . "service_path" "/service/" -}}
+{{- $_ := set . "v2_path" "/v2/" -}}
+{{- $_ := set . "chartrepo_path" "/chartrepo/" -}}
+{{- $_ := set . "controller_path" "/c/" -}}
+apiVersion: traefik.containo.us/v1alpha1
+kind: IngressRoute
+metadata:
+  name: "{{ template "harbor.ingressRoute" . }}"
+  labels:
+{{ include "harbor.labels" . | indent 4 }}
+{{- if $ingressRoute.annotations }}
+  annotations:
+{{ toYaml $ingressRoute.annotations | indent 4 }}
+{{- end }}
+spec:
+  entryPoints:
+{{ toYaml $ingressRoute.entryPoints | indent 2 }}
+  routes:
+  - kind: Rule
+    match: Host(`{{ $ingressRoute.hosts.core }}`)
+    services:
+    - kind: Service
+      name: {{ template "harbor.portal" . }}
+      port: 80
+  - kind: Rule
+    match: Host(`{{ $ingressRoute.hosts.core }}`) && (PathPrefix(`{{ .api_path }}`) || PathPrefix(`{{ .service_path }}`) || PathPrefix(`{{ .v2_path }}`) || PathPrefix(`{{ .chartrepo_path }}`) || PathPrefix(`{{ .controller_path }}`))
+    services:
+    - kind: Service
+      name: {{ template "harbor.core" . }}
+      port: 80
+  {{- if .Values.notary.enabled }}
+  - kind: Rule
+    match: Host(`{{ $ingressRoute.hosts.notary }}`)
+    services:
+    - kind: Service
+      name: {{ template "harbor.notary-server" . }}
+      port: 4443
+  {{- end }}
+{{- if $ingressRoute.tls }}
+  tls: 
+{{ toYaml $ingressRoute.tls | indent 4 }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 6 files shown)


MatthiasHertel:
can we merge that ?

rbtr:
that would be great

rbtr:
@ywk253100 does this have any potential to be merged?

ywk253100:
Sorry for the late reply.
@rbtr May I know more about the background? Seem that you got error when you deploy Harbor chart with Traefik ingress controller because there are two proxies: Traefik and Nginx. But the Nginx pod will not be deployed if you choose the ingress as the way to expose Harbor service

rbtr:
@ywk253100 yes you're correct, I had errors getting Harbor to work when Nginx and Traefik are both proxying traffic. In my cluster, Traefik is doing SSL termination and that seems to be part of the problem (Traefik has very good integration with [LetsEncrypt](https://letsencrypt.org/) and I think this is a very common set-up).

But just setting the type to `Ingress` in the Harbor helm chart is not enough. To be recognized by  Traefik, it requires the `IngressRoute` configuration - a Traefik CRD that has more features than the standard `Ingress`.

This PR adds the IngressRoute as a configuration options so that this helm chart can be deployed directly in to clusters that use Traefik for the Ingress Controller.

Does that clear things up?

templates/ingress/ingressroute.yaml
```diff
@@ -0,0 +1,43 @@
...
+      port: 80
+  {{- if .Values.notary.enabled }}
+  - kind: Rule
+    match: Host(`{{ $ingressRoute.hosts.notary }}`)
+    services:
+    - kind: Service
+      name: {{ template "harbor.notary-server" . }}
+      port: 4443
+  {{- end }}
+  tls:
```
stephenwolfe1: Might want to wrap `tls:` in a conditional so the rendered file does not have it listed when tls is empty. Traefik still reads the empty object and thinks tls is enabled on the router which does not work if you are terminating at an external load balancer.

rbtr: thanks for that correction 👍

sulaweyo:
Any news on this? I just ran into this issue with traefik as ingress

rbtr:
I'm keeping my branch up to date with that latest master :man_shrugging:

Tchoupinax:
Hello :)

What's up with this branch ? ^^

rodrigorras (approved):

hgye:
I'm also face the same issue: try using traefik ingressroute instead of ingress. 
I'm using clusterIP instead of ingress.
Then connect ingressroute to service harbor-portal.

I can get harbor login page, but I try to login with default user/pass(admin/Harbor12345), it says "invalid username or password"

Is that possible using traefik ingressroute for harbor??

following is my ingressroute yaml:
```yaml=
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  name: harbor-ingress
  namesapce: harbor
spec:
  entryPoints:
    - websecure
  routes:
  - match: Host(`harbor.example.io`)
    kind: Rule
    services:
    - name: harbor-harbor-portal
      port: 80
    middleware:
    - name: limit
      namespace: harbor
  tls:
    secretName: harbor-onuware-io-tls
    # domains:
    # - main: YOUR-DOMAIN.com.cn
    #   sans:
      # - YOUR-APP.YOUR-DOMAIN.com.cn
    # certResolver: default
    # options: {}

---
apiVersion: traefik.containo.us/v1alpha1
kind: Middleware
metadata:
  name: limit
  namespace: harbor
spec:
  buffering:
    memRequestBodyBytes: 6000000000 #  6G
```

Tchoupinax:
Hello @hgye !

I use this ingressRoute and It works well :)

```yaml
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  name: harbor-tls
  annotations:
    prometheus.io/scrape: 'true'
spec:
  entryPoints:
    - websecure
  tls:
    secretName: harbor-cert
  routes:
  - match: Host(`harbor.mydomain.com`)
    kind: Rule
    services:
    - name: harbor-harbor-portal
      port: 80
  - match: Host(`harbor.mydomain.com`) && PathPrefix(`/api/`, `/service/`, `/v2/`, `/chartrepo/`, `/c/`)
    kind: Rule
    services:
    - name: harbor-harbor-core
      port: 80
```

FYI, I discovered that Traefik also understand classic ingress. You have the arg `--providers.kubernetesingress`. For this. I did not try because my ingressRoute does the job but maybe you could find this useful. ^^
More details here : https://github.com/traefik/traefik-helm-chart

Have a nice day

amrap030:
Hello together, how would you do this, if you want to server harbor via a different path. When I add an ingress for `https://<hostname>/harbor` the first request from the browser is fine, then it shows loading... on the screen, but nothing happens. The reason is, that all subsequent requests are being made against `https://<hostname>/` without the path. Is it possible to set something like a base or root url for harbor?

Issue #408, opened 2019-10-21
accessMode ReadWriteOnce blocks upgrade because of mount already mounted


roflmao:
Harbor 1.9.1 from Helm.

ReadWriteOnce with default strategy RollingUpdate blocks upgrade since the volume is already mounted.

I fixed it locally with changing the deployment to:

```
  strategy:
    type: Recreate
```

tlvenn:
@ywk253100 any chance we can add strategy.type value for all the deployments that are using a pvc ? The default value should probably be `Recreate` as I expect the vast majority of users will have only `RWO` pvc outside of the box.

tlvenn:
This [PR](https://github.com/goharbor/harbor-helm/pull/366) would solve the issue.
@ywk253100 can you please review it ? Thanks

ywk253100:
Fixed

PR #470, opened 2019-12-16
Provide the update strategy for deployments with persistent volumes in values.yaml

### LINKED ISSUE #224: Harbor helm upgrade fails due to volume multi attach error on multiple workers

ywk253100:
Fixes #224, #408

With the default update strategy(RollingUpdate), the pods are stuck for waiting volumes if the volumes don't support RWM. This commit exposes the configuration item, so that users can modify it when needed


README.md
```diff
@@ -160,8 +160,9 @@ The following table lists the configurable parameters of the Harbor chart and th
...
-| `imagePullSecrets`                                                          | The imagePullSecrets names for all deployments                                                                                                                                                                                                                                                                                                  |                                 |
+| `imagePullPolicy` | The image pull policy |  |
+| `imagePullSecrets` | The imagePullSecrets names for all deployments |  |
+| `updateStrategy.type` | The update strategy for deployments with persistent volumes(jobservice, registry and chartmuseum): `RollingUpdate` or `Recreate`. Setting it as `Recreate` when `RWM` for volumes isn't supported  | `RollingUpdate` |
```
reasonerjt: Setting -> Set

values.yaml
```diff
@@ -240,6 +240,12 @@ imagePullSecrets:
 #  - name: docker-registry-secret
 #  - name: internal-registry-secret
 
+# The update strategy for deployments with persistent volumes(jobservice, registry 
+# and chartmuseum): "RollingUpdate" or "Recreate"
+# Setting it as "Recreate" when "RWM" for volumes isn't supported
```
reasonerjt: Setting -> set

Issue #480, opened 2019-12-31
jobservice CrashLoopBackOff  -- redis volume permission issue.


ruanshudong:
tars-harbor-redis is running, but jobservice is CrashLoopBackOff.

kubectl logs tars-harbor-jobservice-578f7b68bd-8nv5f

2019-12-31T01:55:00Z [DEBUG] [/common/rbac/casbin.go:97]: Starting regexp store purge in 1m0s
2019-12-31T01:55:00Z [INFO] [/replication/adapter/native/adapter.go:42]: the factory for adapter docker-registry registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/harbor/adapter.go:40]: the factory for adapter harbor registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/dockerhub/adapter.go:25]: Factory for adapter docker-hub registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/huawei/huawei_adapter.go:27]: the factory of Huawei adapter was registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/googlegcr/adapter.go:29]: the factory for adapter google-gcr registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/awsecr/adapter.go:47]: the factory for adapter aws-ecr registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/azurecr/adapter.go:15]: Factory for adapter azure-acr registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/aliacr/adapter.go:26]: the factory for adapter ali-acr registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/jfrog/adapter.go:30]: the factory of jfrog artifactory adapter was registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/quayio/adapter.go:34]: the factory of Quay.io adapter was registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/helmhub/adapter.go:30]: the factory for adapter helm-hub registered
2019-12-31T01:55:00Z [INFO] [/replication/adapter/gitlab/adapter.go:19]: the factory for adapter gitlab registered
2019-12-31T01:55:00Z [INFO] [/common/config/store/driver/rest.go:24]: get configuration from url: http://tars-harbor-core/api/internal/configurations
[...]

Issue #379, opened 2019-10-02
Web interface login broken


jeandevops:
Hello! I'm trying to use the helm chart (harbor-1.2.0 - app version 1.9.0) at Kubernetes 1.14.6, but I'm facing the same problem described at the issues:

https://github.com/goharbor/harbor/issues/4161
https://github.com/goharbor/harbor/issues/3418

For the first time, I can login and change the password. I even can login again one time or two, but after some time the login button brokes.

I'm hitting the web interface directly through Nginx component (no Ingress or proxy in the way), for this I've made some customizations at Nginx template deployment.yaml to use "hostNetwork" and "ClusterFirstWithHostNet" _(because we use Kong as Ingress controller and Harbor brokes the authentication headers used by OpenID at registry - that's a problem related at issue #3114 at goharbor/harbor with "X-Forwarded-Proto $scheme", I've tried to follow the workarounds but had none success - but this is a talk for another day ^^)_

That's a piece of log from the core pod:

```
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:227]: OIDC CLI modifier only handles request by docker CLI or helm CLI
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:442]: can not get user information from session
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:511]: user information is nil
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:525]: using local database project manager
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:527]: creating local database security context...
2019-10-01T19:10:01Z [DEBUG] [/common/dao/user.go:269]: Check if user admin is super user
2019-10-01T19:10:01Z [DEBUG] [/core/auth/authenticator.go:139]: Current AUTH_MODE is db_auth
2019/10/01 19:10:01 [D] [server.go:2774] |  172.18.12.232| 200 |  12.986366ms|   match| POST     /c/login   r:/c/login
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:227]: OIDC CLI modifier only handles request by docker CLI or helm CLI
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:442]: can not get user information from session
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:511]: user information is nil
2019-10-01T19:10:01Z [DEBUG] [/core/filter/security.go:525]: using local database project manager
[...]

PR #484, opened 2020-01-08
Add redis init container to change permission of directory

secret104278:
If permission of redis data volume not correct, redis will not work, and other service depend on redis will fail to start(e.g. jobservice). Furthermore, if jobservice fail to start, web GUI login may fail on higher version of harbor(e.g. v1.10.0).

fixed issues: https://github.com/goharbor/harbor-helm/issues/480, https://github.com/goharbor/harbor-helm/issues/379


secret104278:
This patch should be added to `1.20` and `1.30`, because harbor `v1.9.4` and `v1.10.0` web GUI login will fail due to unhealthy redis (see https://github.com/goharbor/harbor-helm/issues/379#issuecomment-570032165)

ywk253100:
@secret104278 The permission issue should be covered by security context: https://github.com/goharbor/harbor-helm/blob/master/templates/redis/statefulset.yaml#L28

secret104278:
@ywk253100 
I see. I found that my issue was cause by default `localPath` storageClass which fsGroup will fail to change volume permission.

PR #509, opened 2020-02-13
Support externally-managed serviceAccounts for registry and chartmuseum for IRSA

jonathan-mothership:
When running Harbor on EKS and using S3 for Registry and Chartmuseum there is an opportunity to manage AWS credentials using "IAM Roles for Service Accounts" (IRSA). The upstream Registry and Chartmuseum projects don't yet have working IRSA support but it's just around the corner! I've tested this PR against our forks of those projects (using Harbor's build process for them) and using Service Accounts managed outside of helm I'm able to use AWS S3 as the storage provider without explicit credentials or running KIAM/Kube2IAM.


alex1989hu:
I think this can be closed due to https://github.com/goharbor/harbor-helm/pull/573 👍

PR #519, opened 2020-02-21
Add Trivy scanner

danielpacak:


values.yaml
```diff
@@ -447,6 +453,40 @@ clair:
   ## Additional deployment annotations
   podAnnotations: {}
 
+trivy:
+  # enable is the flag to enable Trivy scanner
+  enabled: true
```
ywk253100: Both trivy and clair are enabled by default, is this by design?

templates/trivy/trivy-sts.yaml
```diff
@@ -0,0 +1,102 @@
+{{- $trivyPVC := .Values.persistence.persistentVolumeClaim.trivy -}}
+apiVersion: apps/v1
+kind: StatefulSet
```
ywk253100: Why it's `StatefulSet`? Isn't trivy a stateless application just like clair?

values.yaml
```diff
@@ -447,6 +453,40 @@ clair:
...
+  # replicas the number of Pod replicas
+  replicas: 1
+  # cacheDir is Trivy cache directory
+  # This is where Trivy stores its vulnerabilities DB/index and caches Docker images. The size of the cache might grow
+  # over time and requires housekeeping.
+  cacheDir: "/home/scanner/.cache/trivy"
+  # reportsDir is Trivy reports directory
+  # It's used by the scanner adapter to store a temporary scan report generated by Trivy in JSON format before it's
+  # transformed and stored in Harbor DB.
+  reportsDir: "/home/scanner/.cache/reports"
```
ywk253100: Seems that configuring the `cacheDir` and `reportsDir` for users is meaningless as they are both the directories inside the container

templates/trivy/trivy-sts.yaml
heww: Trivy use file system to save its db and reports.

values.yaml
steven-zou: I think it should be Trivy and Clair can be both deployed as default scanners. Under this situation, Trivy is the default, Clair is not.

For enablement, I think Trivy should be enabled as 1st choice and Clair as a second choice which needs users to explicitly specify `enabled:true`.

heww: Both trivy and clair are enabled, trivy as default.

templates/trivy/trivy-sts.yaml
danielpacak: That's correct. As @heww mentioned Trivy adapter is currently stateful as explained here https://github.com/aquasecurity/harbor-scanner-trivy/blob/master/docs/ARCHITECTURE.md

However, we're working on so called [Trivy server - client mode](https://github.com/aquasecurity/trivy#client--server). If we take advantage of that we could go for Deployment instead of StatefulSet.

values.yaml
danielpacak: That's a good point. I'll remove it.

templates/trivy/trivy-sts.yaml
danielpacak: And to be precise Clair is stateful as well. It stores it state in Postgres DB.

PR #523, opened 2020-02-29
Edited log level comments

utkudarilmaz:
When logLevel set to warning as described in comment on values.yaml, registry container failing with _'Invalid logLevel warning Must be one of [error, warn, info, debug]'_ error. Also same error occuring when logLevel setted to 'fatal'.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -246,7 +246,7 @@ imagePullSecrets:
 updateStrategy:
   type: RollingUpdate
 
-# debug, info, warning, error or fatal
+# debug, info, warn or error
 logLevel: info
 
 # The initial password of Harbor admin. Change it from portal after launching Harbor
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -246,7 +246,7 @@ imagePullSecrets:
 updateStrategy:
   type: RollingUpdate
 
-# debug, info, warning, error or fatal
+# debug, info, warn or error
```
ywk253100, line 249: @utkudarilmaz Thank you for pointing out this.
The values of log level are different for different components.
The `debug`, `info`, `warning`, `error` and `fatal` work for `core` and the registry only supports `debug`, `info`, `warn` and `error`. So I think we can do some convert for registry

reasonerjt:
So this is not needed once #704 is merged, right?

ywk253100:
> So this is not needed once #704 is merged, right?

Yes

PR #531, opened 2020-03-13
feat: Allow configuring SCANNER_TRIVY_GITHUB_TOKEN env

danielpacak:


templates/trivy/trivy-secret.yaml
```diff
@@ -7,5 +7,6 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
 type: Opaque
 data:
-  redis: {{ include "harbor.redisForTrivyAdapter" . | b64enc }}
+  redisURL: {{ include "harbor.redisForTrivyAdapter" . | b64enc }}
+  gitHubToken: {{ .Values.trivy.gitHubToken | quote }}
```
ywk253100: Isn't this token needed to be encoded with base64?

steven-zou: @danielpacak 

What do you think?

danielpacak: Yes, I think I have missed that. Let me double check if that's the case I'll open a small PR to fix it. Thanks @ywk253100 for pointing this out. Good catch!

PR #535, opened 2020-03-17
Switch auth mode of registry, and other cleanup

reasonerjt:
This commit makes change to the chart aligning with the update in Harbor
2.0:

1. Switch auth mode of registry from token to htpasswd
2. Remove notification section of registry config
3. Disable XSRF from beego


values.yaml
```diff
@@ -395,6 +395,13 @@ registry:
   secret: ""
   # If true, the registry returns relative URLs in Location headers. The client is responsible for resolving the correct URL.
   relativeurls: false
+  credentials:
```
ywk253100: Add them to the readme.md

reasonerjt: added

PR #542, opened 2020-03-19
Update the env of registry URL

ywk253100:
Update the evn of registry and controller URL used by core and jobservice


bclermont:
There is 35 pull requests pending reviews, many of them seem very legits and actually improve and fix this chart

The only way to be reviewed and merged is to have a @vmware.com email address?

That is not how community works.

bclermont:
> That is not how community works.

Or maybe the whole point of this chart is to let vmware deploy it? and ignore the users community?

PR #546, opened 2020-03-24
Check trivy.enabled in statefulset

### LINKED ISSUE #545: trivy: do not install if disabled
The StatefulSet is installed even if `.Values.trivy.enabled=false`

alex1989hu:
Do not generate Trivy StatefulSet if it is not enabled

Fixes #545

Signed-off-by: Alex Szakaly <alex.szakaly@gmail.com>

### DIFF
--- a/templates/trivy/trivy-sts.yaml
+++ b/templates/trivy/trivy-sts.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.trivy.enabled }}
 {{- $trivyPVC := .Values.persistence.persistentVolumeClaim.trivy -}}
 apiVersion: apps/v1
 kind: StatefulSet
@@ -105,3 +106,4 @@ spec:
         resources:
           requests:
             storage: {{ $trivyPVC.size | quote }}
+{{- end }}


templates/trivy/trivy-sts.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.trivy.enabled }
```
danielpacak: I think you missed a curly bracket here ;)
```suggestion
{{- if .Values.trivy.enabled }}
```

templates/trivy/trivy-sts.yaml
```diff
@@ -105,3 +106,4 @@ spec:
         resources:
           requests:
             storage: {{ $trivyPVC.size | quote }}
+{{ end }}
```
danielpacak: We can trim left whitespaces here:

```suggestion
{{ end }}
{{- end }}
```

templates/trivy/trivy-sts.yaml
alex1989hu: Oh my, sorry .. :) 👍

alex1989hu:
@danielpacak: Updated and rebased!

Thank you, it was early in the morning, nice catch too!

PR #552, opened 2020-03-29
Use postgresql as chart dependency

lwj5:
Hi,

This PR helps to:
- reduce the need to maintain extra components in the chart
- maintain one less docker image
- reduce taking on new issues arising from PostgreSQL problems such as root user and PSP
- enjoy all the new features such as the replication and other configurations that are not available in this chart

This allows Harbor and its contributors to focus its effort on developing the core and shift the development of dependencies elsewhere.

I have proposed this in slack some time ago in a chat with Steven
https://cloud-native.slack.com/archives/CC1E0J0MC/p1576670854007800

Since now I have some time and the helm dependencies are decentralised. Here's the PR :)

If we need to plan for some form of deprecation or switch over feel free to let me know


yangwang888:
yes,we need this ,the postgres is not stable at all

values.yaml
```diff
@@ -535,27 +527,24 @@ notary:
...
+  postgresqlPassword: changeit
+  postgresqlDatabase: registry
+  initdbScripts:
+    db-init.sql: |
+      CREATE DATABASE clair;
+      CREATE DATABASE notary_server;
+      CREATE DATABASE notary_signer;
+      GRANT ALL PRIVILEGES ON DATABASE clair TO harbor;
+      GRANT ALL PRIVILEGES ON DATABASE notary_server TO harbor;
+      GRANT ALL PRIVILEGES ON DATABASE notary_signer TO harbor;
```
MrD2, line 543: why not have this in the databse sub-section?

lwj5: Because this chart variation depends on the postrgresql chart as listed in requirements.yaml.
Any values affecting the sub-chart will need to be prefixed by the chart name.

lwj5:
I'm closing this PR in favour of https://github.com/bitnami/charts/tree/master/bitnami/harbor which is a more stable and secure platform. This place doesnt seem to value community inputs.

PR #555, opened 2020-04-01
Add redis init container to change permission of directory

secret104278:
If permission of redis data volume not correct, redis will not work, and other service depend on redis will fail to start(e.g. jobservice). Furthermore, if jobservice fail to start, web GUI login may fail on higher version of harbor(e.g. v1.10.0).

Although `fsGroup` is set, but k8s only support certain types of fs to apply the permission. Use init container will make this chart work in most senario.

fixed issues: https://github.com/goharbor/harbor-helm/issues/480, https://github.com/goharbor/harbor-helm/issues/379


secret104278:
Can Any one review this?

PR #572, opened 2020-04-17
feat(feature) allow define external registry instead of docker hub

bitsf:
allow define external registry instead of docker hub

--set externalReg=vmware.io/goharbor/

Signed-off-by: Ziming Zhang <zziming@vmware.com>


values.yaml
```diff
@@ -365,7 +366,7 @@ jobservice:
 registry:
   registry:
     image:
-      repository: goharbor/registry-photon
+      repository: registry-photon
```
reasonerjt, line 369: why not just let user set the repo with registry's url?

bitsf: we have so many repos, user don't need to set each one, and that would make the wrapper scripts too complex.
My target is just push the harbor images to some different registries.

reasonerjt: I don't think we should trade flexibility for a shortcut.   
For example, when I'm debugging the chart I can push an image to my local registry or my personal namespace and use official images for other pods.

If you don't like editing the values.yaml, you can write a script or use tools like kustomize to patch the yaml?

reasonerjt:
Per discussion with @bitsf offline, we agree this is not the right way, hence I'm closing this PR.

PR #575, opened 2020-04-19
Add internal config for chart

ninjadq:


templates/chartmuseum/chartmuseum-dpl.yaml
```diff
@@ -70,17 +80,42 @@ spec:
           - name: NO_PROXY
             value: "{{ template "harbor.noProxy" . }}"
           {{- end }}
+          {{- if .Values.internalTLS.enabled }}
```
ywk253100, line 83: What are these environment variables used for?

values.yaml
```diff
@@ -100,6 +100,91 @@ expose:
...
 
+# The internal TLS used for harbor components secrue communicating. In order to enable https 
+# in each components tls cert files need to provided in advance.
+internalTLS:
+  # If internal TLS enabled
+  enabled: false
+  # There are three ways to provide tls
+  # 1) "auto" will generate cert automatically
+  # 2) "manual" need provide cert file manually in following value
+  # 3) "cert-manager" will manage cert file by cert-manager
```
ywk253100, line 111: Is this a valid option? What happens if set `certSource` as `cert-manager`?

values.yaml
```diff
@@ -100,6 +100,91 @@ expose:
...
+  certSource: "auto"
+  # The content of trust ca, only available when `certSrouce` is "manual"
+  trustCa: ""
+  # core related cert congfiguration
+  core:
+    # secret name for core's tls certs
+    secretName: "harbor-core-internal-tls"
```
ywk253100, line 118: The value is also used as the name of secret if I set the `certSource` as `auto`, this will cause that I cannot deploy more than one Harbor instance in the same namespace

values.yaml
```diff
@@ -23,7 +23,7 @@ expose:
     notarySecretName: ""
     # The common name used to generate the certificate, it's necessary
     # when the type isn't "ingress" and "secretName" is null
-    commonName: ""
+    commonName: "core.harbor.domain"
```
ywk253100, line 26: This shouldn't be set

values.yaml
```diff
@@ -100,6 +100,91 @@ expose:
 # If Harbor is deployed behind the proxy, set it as the URL of proxy
 externalURL: https://core.harbor.domain
 
+# The internal TLS used for harbor components secrue communicating. In order to enable https
```
ywk253100, line 103: secure is typo

values.yaml
```diff
@@ -100,6 +100,91 @@ expose:
 # If Harbor is deployed behind the proxy, set it as the URL of proxy
 externalURL: https://core.harbor.domain
 
+# The internal TLS used for harbor components secrue communicating. In order to enable https 
+# in each components tls cert files need to provided in advance.
+internalTLS:
```
ywk253100, line 105: Please add all the new items in the `README.md` as well

templates/core/core-cm.yaml
```diff
@@ -36,20 +43,36 @@ data:
...
   CHART_REPOSITORY_URL: "http://{{ template "harbor.chartmuseum" . }}"
+{{- end }}
   LOG_LEVEL: "{{ .Values.logLevel }}"
   CONFIG_PATH: "/etc/core/app.conf"
   SYNC_REGISTRY: "false"
   CHART_CACHE_DRIVER: "redis"
   _REDIS_URL: "{{ template "harbor.redisForCore" . }}"
   _REDIS_URL_REG: "{{ template "harbor.redisForGC" . }}"
   PORTAL_URL: "http://{{ template "harbor.portal" . }}"
+{{- if .Values.internalTLS.enabled }}
```
ywk253100, line 71: How about move all these checking for internal tls in the same `if else end` block?

templates/chartmuseum/chartmuseum-cm.yaml
```diff
@@ -6,7 +6,15 @@ metadata:
   labels:
 {{ include "harbor.labels" . | indent 4 }}
 data:
+{{- if .Values.internalTLS.enabled }}
+  PORT: "9443"
+  TLS_CERT: "/etc/harbor/ssl/chartmuseum.crt"
+  TLS_KEY: "/etc/harbor/ssl/chartmuseum.key"
+{{- else }}
   PORT: "9999"
+  TLS_CERT: ""
```
reasonerjt, line 15: TLS_CERT and TLS_KEY are not needed for serving HTTP.

values.yaml
reasonerjt, line 111: When certSource is set as "cert-manager", what do you do?

heww: It was changed to `secret`

values.yaml
heww: secretName for `auto` and `manual` will be `{{ release-name }}-xxx`

values.yaml
heww: It was renamed to `secret`.

templates/chartmuseum/chartmuseum-dpl.yaml
heww: It was used to let chartmuseum run in https mode.

values.yaml
```diff
@@ -100,6 +100,75 @@ expose:
...
+  # If internal TLS enabled
+  enabled: false
+  # There are three ways to provide tls
+  # 1) "auto" will generate cert automatically
+  # 2) "manual" need provide cert file manually in following value
+  # 3) "secret" internal certificates from secret
+  certSource: "auto"
+  # The content of trust ca, only available when `certSource` is "manual"
+  trustCa: ""
+  # core related cert congfiguration
```
ywk253100, line 115: congfiguration -> configuration

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 1.2.0-dev
+version: 1.4.0-dev
```
renmaosheng, line 3: do we need to bump it to 2.0.0? I know our helm version is different from Harbor OSS version.

README.md
```diff
@@ -126,6 +126,31 @@ The following table lists the configurable parameters of the Harbor chart and th
 | `expose.loadBalancer.ports.notaryPort` | The service port Notary listens on. Only needed when `notary.enabled` is set to `true`|
 | `expose.loadBalancer.annotations` | The annotations attached to the loadBalancer service | {} |
 | `expose.loadBalancer.sourceRanges` | List of IP address ranges to assign to loadBalancerSourceRanges | [] |
+| **Internal TLS** |
+| `internalTLS.enabled` | Enable the tls for the components (chartmuseum, clair, core, jobservice, portal, registry, trivy) | `false` |
+| `internalTLS.certSource` | Method to provide tls for the components, options is `auto`, `manual`, `secret` | `auto` |
+| `internalTLS.trustCa` | The content of trust ca, only available when `certSrouce` is `manual` |  |
```
reasonerjt, line 132: this trust CA should be the CA that signs the internal certificate correct? could you clarify that in the doc?
Please also elaborate what the secret should look like when the certsource is set to `secret`

Chart.yaml
tvvignesh: @renmaosheng Was wondering the same till I saw ur comment :smile: So, this is for 2.0.0 right?

PR #631, opened 2020-06-17
Update Ingress apiVersion for support Kubernetes v1.16

### LINKED ISSUE #628: Replace "extensions/v1beta1" to use "networking.k8s.io/v1beta1" on Ingress Template to support Kubernetes v1.16
This issue is different from this one:

- [migrate to use apps/v1 for support kubernetes 1.16 #403](https://github.com/goharbor/harbor-helm/issues/403)

Because that issue does not consider **extensions/v1beta1** API deprecated usage found here:

1. https://github.com/goharbor/harbor-helm/blob/v1.4.0/templates/ingress/ingress.yaml#L28
2. https://github.com/goharbor/harbor-helm/blob/master/templates/ingress/ingress.yaml#L91

This `apiVersion` should be replaced for `networking.k8s.io/v1beta1` as described [here](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/) to support Kubernetes v1.16.

Actually I didn't identify a specific issue that breaks the functionality of this Chart, but that API version should be replaced for full support Kubernetes v1.16.

ramrodo:
This PR is for update Ingress apiVersion from `extensions/v1beta1` to `networking.k8s.io/v1beta1` for supporting Kubernetes v1.16.

Closes #628 

Signed-off-by: Rodolfo Martínez <rodomar@outlook.com>

### DIFF
--- a/templates/ingress/ingress.yaml
+++ b/templates/ingress/ingress.yaml
@@ -25,7 +25,7 @@
 {{- end }}
 
 ---
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: "{{ template "harbor.ingress" . }}"
@@ -88,7 +88,7 @@ spec:
 
 {{- if .Values.notary.enabled  }}
 ---
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: "{{ template "harbor.ingress-notary" . }}"
--- a/templates/ingress/ingress.yaml
+++ b/templates/ingress/ingress.yaml
@@ -25,11 +25,11 @@
 {{- end }}
 
 ---
-{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.Version -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.Version }}
 apiVersion: extensions/v1beta1
-{{- else -}}
+{{- else }}
 apiVersion: networking.k8s.io/v1beta1
-{{- end -}}
+{{- end }}
 kind: Ingress
 metadata:
   name: "{{ template "harbor.ingress" . }}"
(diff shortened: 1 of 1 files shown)


heww:
Please use [Capabilities](https://helm.sh/docs/chart_template_guide/builtin_objects/) to deal with this problem when both helm 2 and 3 support `Capabilities` feature. Thanks!

ramrodo:
Thanks for the tip @heww. I just updated the PR.

maxdanilov:
Hi 👋 
Are there any updates on this?
We were wondering why the 1.4.1 chart is using a deprecated ingress version and found this PR.
Any estimation on when it will be released?

ryan-a-baker:
Can we get this merged?  This is blocking us from upgrading to 1.16.

stefannica (approved):
Thanks, I recently ran into this problem myself, tried to fix it, but didn't see this PR was already open.

ramrodo:
Thanks Stefan. There is pending one more review in order to merge this. Can anyone else help us with this @reasonerjt or @ywk253100?

jenting (approved):

ryan-a-baker:
It was just pointed out to me that this actually shouldn't block anything on 1.16 (I tested and confirmed as well on 1.16).  I was just assuming that it would block us because it came back on the [mapkubeapis](https://github.com/hickeyma/helm-mapkubeapis) plugin as having a deprecated API.  Ingresses aren't removed in 1.16, but instead in 1.22, which is what this PR resolves.

```
The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

Ingress in the extensions/v1beta1 API version will no longer be served
Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.
```

https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16

Still would be good to get this merged, but isn't blocking on 1.16 upgrades as I mentioned previously.

reasonerjt (approved):

reasonerjt:
Thanks and I'll write another PR to remove the `if` chunks as we'll only support v1.16+ since v1.5.0 chart.

ywk253100 (approved):

PR #647, opened 2020-07-01
Fix the filesystem permission for registry and redis using initContainer

huats:


templates/registry/registry-dpl.yaml
```diff
@@ -44,6 +44,16 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:                                                           
+      - name: "change-permission-of-directory"                                  
+        image: {{ .Values.registry.registry.initContainerImage.repository }}:{{ .Values.registry.registry.initContainerImage.tag }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}                          
+        command: ["/bin/sh"]                                                    
+        args: ["-c", "chown -R 10000:10000 /storage"]
```
stefannica, line 52: Please don't hardcode the `/storage` location here. You should use the values.yaml parameter value `{{ .Values.persistence.imageChartStorage.filesystem.rootdirectory }}` instead.
You can also use `harbor:harbor` instead of `10000:10000` if you reuse the `harbor-registry` or `harbor-registryctl` image.

values.yaml
```diff
@@ -455,6 +455,9 @@ registry:
     #  requests:
     #    memory: 256Mi
     #    cpu: 100m
+    initContainerImage:
```
stefannica, line 458: You don't need a different init container image. You should simply reuse the `harbor-registry` or `harbor-registryctl` container image. This way, you also have the harbor username and group name, which means you can reference them directly instead of using the hardcoded UID and GID values.

values.yaml
```diff
@@ -700,6 +703,9 @@ redis:
     image:
       repository: goharbor/redis-photon
       tag: dev
+    initContainerImage:
```
stefannica, line 706: You don't need a different init container image. You should simply reuse the `harbor-redis` container image. This way, you also have the redis username and group name, which means you can reference them directly instead of using the hardcoded UID and GID values.

templates/redis/statefulset.yaml
```diff
@@ -33,6 +33,18 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+{{- if eq .Values.redis.type "internal" }}                                      
+      initContainers:                                                           
+      - name: "change-permission-of-directory"                                  
+        image: {{ .Values.redis.internal.initContainerImage.repository }}:{{ .Values.redis.internal.initContainerImage.tag }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}                          
+        command: ["/bin/sh"]                                                    
+        args: ["-c", "chown -R 999:999 /var/lib/redis"]
```
stefannica, line 42: You can also use `redis:redis` instead of `999:999`, if you reuse the `harbor-redis` image.

stefannica (changes requested):
Thanks, this is a good idea, but you can take it one step further, with little effort:

* instead of using a busybox image for the init container, you should simply reuse the container image used for the actual workload (see this PR for a similar example: #685).
* don't use hardcoded values, if possible. You can even replace UID and GID values with the actual username and group name, if you reuse the container image as suggested above.

reasonerjt:
Why `fsGroup` does not work for you?

huats:
During my differents tests fsGroup was not working correctly probably due to the type of storage I had (a PV/PVC with underlying NFS that is created using rook)

huats:
If you think there is an interest, I'd be really happy to apply all the good comments from @stefannica stefannica

stefannica:
> If you think there is an interest, I'd be really happy to apply all the good comments from @stefannica

@huats there is definitely interest from me :-)

ywk253100:
The initContainer solution doesn't work in the restricted psp enabled environment. 
Closing this PR

PR #493, opened 2020-01-17
fix bugs that postgres can't start

sunlianqiang:
after k8s node restart, the postgres of harbor can't restart. It's file mode issue when checking logs
```
kubectl logs -f harbor-sh-harbor-database-0 -n harbor-sh
FATAL:  data directory "/var/lib/postgresql/data" has group or world access
DETAIL:  Permissions should be u=rwx (0700).
```
I find a method that change the config of initContainers to fix the bug. I Also found in https://github.com/helm/charts Postgres are configured with master-slave, hoping to use asap,  I also can help if is needed.

FYI:
https://github.com/ClusterHQ/dvol/issues/45
https://github.com/helm/charts/tree/master/stable/postgresql

PR #661, opened 2020-07-08
Postgres: chmod 0700 database-files before startup

msiebuhr:
Fixes the following error on database restart (possibly related to
particulars of PV/PVC setup):

    FATAL:  data directory "/var/lib/postgresql/data" has group or world access
    DETAIL:  Permissions should be u=rwx (0700).


ywk253100:
Dup of https://github.com/goharbor/harbor-helm/pull/493

PR #667, opened 2020-07-14
Run change-permission-of-directory container with DB image

jsuchome:
If the init container is run with the DB image and explicit runAsUser: 0
security context, it has several advantages:

- no need to need extra image for init container (busybox in this case)
- possibility to run "chown" command with user:group name instead of UID:GID,
which means the whole DB deployment can work with different database images,
without the need to adapt UID:GID values


heww:
@jsuchome do you ensure that the user in harbor-db image always has the permission to change the owner of `var/lib/postgresql/data`?

jsuchome:
That's achieved by explicit `runAsUser: 0` for the init container. So even though the original user for such container is  postgres, it's run as root in this case.

ywk253100:
@jsuchome Could you cherry pick the PR to branch 1.4.0 as well?

jsuchome:
Sure, here it is: https://github.com/goharbor/harbor-helm/pull/685

stefannica:
@reasonerjt could you please approve this PR too ? the 1.4.0 backport was already merged.

Anticom:
Am I missing something? I've installed 1.4.0 explicitly but `change-permission-of-directory` Init Container still fails with message `chown: /var/lib/postgresql/data: Operation not permitted` and upon inspecting the deployed init container there's no trace of
```yaml
        securityContext:
          runAsUser: 0
```
to be found. When is this issue expected to be fixed for more recent releases? This is really holding me back from using the chart since the init containers are unconditionally deployed.

jsuchome:
@Anticom Hm, here https://github.com/goharbor/harbor-helm/blob/1.4.0/templates/database/database-ss.yaml#L37 I can see the new security context set for the init container, are you sure you have helm chart from 1.4.0 branch?

Anticom:
This is what I'm seeing on my end: [asciicast](https://asciinema.org/a/yMCCGZxhAnXDHcVxvfzihJ8mf)

Call me out if I'm missing something :wink:

jsuchome:
Well, that's something different. I admit I do not know how `--version` works for helm charts but I would expect it needs some kind of release first. And the mere fact that the code is in related github branch does not necessary mean that ... 

I may be wrong of course. But anyway I do not think _this_ PR (which targets master branch) is the right place for the discussion. Either pick the PR into the 1.4.0 branch or related issue...

stefannica:
@Anticom the fix was backported to the 1.4.0 _github branch_, but it was not released with the 1.4.0 helm chart version. It only made it in 1.4.2.

Anticom:
Thanks @stefannica , that's the one that confused me!

reasonerjt (approved):

PR #659, opened 2020-07-07
add trivy nodeselector and tolerations

jsaulalonso:

PR #676, opened 2020-07-27
Trivy: adds nodeSelector, afinity and tolerations

caruccio:


caruccio:
Closing in favor of #659

PR #677, opened 2020-07-28
Update the max open connection to 1000

ywk253100:
Update the max open connection to 1000

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -685,8 +685,8 @@ database:
   maxIdleConns: 50
   # The maximum number of open connections to the database.
   # If it <= 0, then there is no limit on the number of open connections.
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -685,8 +685,8 @@ database:
   maxIdleConns: 50
   # The maximum number of open connections to the database.
   # If it <= 0, then there is no limit on the number of open connections.
-  # Note: the default number of connections is 100 for postgre.
-  maxOpenConns: 100
+  # Note: the default number of connections is 1024 for postgre of harbor.
```
reasonerjt, line 688: comment says 1024 but the value is 1000

PR #678, opened 2020-07-28
Update the max open connection to 1000

ywk253100:
Update the max open connection to 1000

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -687,8 +687,8 @@ database:
   maxIdleConns: 50
   # The maximum number of open connections to the database.
   # If it <= 0, then there is no limit on the number of open connections.
-  # Note: the default number of connections is 100 for postgre.
-  maxOpenConns: 100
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -687,8 +687,8 @@ database:
   maxIdleConns: 50
   # The maximum number of open connections to the database.
   # If it <= 0, then there is no limit on the number of open connections.
-  # Note: the default number of connections is 100 for postgre.
-  maxOpenConns: 100
+  # Note: the default number of connections is 1024 for postgre of harbor.
+  maxOpenConns: 1000
```
reasonerjt, line 691: comment says it's 1024

heww: 1024 is the max connections for the harbor-db,  set harbor-core max use 1000 connections so that to keep some connections for others, eg, health check in harbor-db.

PR #682, opened 2020-07-31
Update Harbor chart to support redis sentinel

ywk253100:
Update Harbor chart to support redis sentinel


lukasmrtvy:
Seems that there is a problem with sprig function `ternary` in helm2 build task.

I am not sure what exact version of helm2 is used, but:
> ternary was added in sprig 2.15 

Which is helm >=2.10 https://github.com/helm/helm/blob/release-2.10/glide.yaml#L15

lukasmrtvy:
It would be nice to also support `rediss://` scheme.

bitsf:
> It would be nice to also support `rediss://` scheme.

some extra components in harbor does not support rediss now, so we didn't support this yet

lukasmrtvy:
Is there any PR/Issue ? Something to track.. Thanks

ywk253100:
@lukasmrtvy Thank you for pointing out the helm version issue, we'll update the prerequisites of helm in the doc

reasonerjt (approved):

PR #693, opened 2020-08-07
Added support for gce-internal ingress

btomasini:
Allows harbor to be installed on private GKE cluster with private facing ingress.


reasonerjt:
@btomasini seems all we need to do to support gce-internal ingress is adding annotations?

PR #723, opened 2020-09-07
Fix notary_server & notary_signer auth via certs

initd:
Config notary-server.json contain option "key_algorithm" with a wrong value (in case of auto generate certs) which is the cause of the next error in notary-server logs:

```
{"level":"error","msg":"Trust not fully operational: rpc error: code = 13 desc = connection error: desc = \"transport: x509: certificate signed by unknown authority (possibly because of \\\"crypto/rsa: verification error\\\" while trying to verify candidate authority certificate \\\"harbor-notary-ca\\\")\"","time":"2020-09-07T14:16:11Z"}
```

### DIFF
--- a/conf/notary-server.json
+++ b/conf/notary-server.json
@@ -7,7 +7,7 @@
     "hostname": "{{ template "harbor.notary-signer" . }}",
     "port": "7899",
     "tls_ca_file": "/etc/ssl/notary/ca.crt",
-    "key_algorithm": "ecdsa"
+    "key_algorithm": "rsa"
   },
   "logging": {
     "level": "{{ .Values.logLevel }}"


reasonerjt:
@ywk253100 please double check if this is covered in the CI pipeline.

ywk253100:
@initd Could you provide a more detail steps to reproduce the error? I cannot reproduce the error in my env.

flbla:
I have the same error in my notary-server logs

milossy:
{"level":"error","msg":"Trust not fully operational: rpc error: code = 13 desc = connection error: desc = \"transport: x509: certificate has expired or is not yet valid: current time 2023-11-02T06:10:02Z is after 2023-04-11T08:30:50Z\"","time":"2023-11-02T06:10:12Z"}

Through my preliminary judgment, it is caused by the certificate expiration problem. It is not clear why the certificate expires. I am not sure which domain name address is used for this certificate, so I need to answer this question

PR #732, opened 2020-09-17
Add startup probe to harbor-core

### LINKED ISSUE #502: liveness probe may be problematic in upgrade
Currently the liveness probe of core calls the `ping` api for checking the liveness.
https://github.com/goharbor/harbor-helm/blob/master/templates/core/core-dpl.yaml#L38

Although the initial delay is configurable, the core may do migration work before serving the requests and the time is not easy to predict.  it's possible that the probe may kill the pod when the migration is in progress, which will cause data in consistency.

We should consider other why for the probe.

reasonerjt:
Fixes #502

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ The following table lists the configurable parameters of the Harbor chart and th
 | `core.image.repository` | Repository for Harbor core image | `goharbor/harbor-core` |
 | `core.image.tag` | Tag for Harbor core image | `dev` |
 | `core.replicas` | The replica count  | `1` |
-| `core.livenessProbe.initialDelaySeconds` | The initial delay in seconds for the liveness probe | `300` |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -211,7 +211,7 @@ The following table lists the configurable parameters of the Harbor chart and th
 | `core.image.repository` | Repository for Harbor core image | `goharbor/harbor-core` |
 | `core.image.tag` | Tag for Harbor core image | `dev` |
 | `core.replicas` | The replica count  | `1` |
-| `core.livenessProbe.initialDelaySeconds` | The initial delay in seconds for the liveness probe | `300` |
+| `core.startupProbe.initialDelaySeconds` | The initial delay in seconds for the startup probe | `300` |
```
ywk253100, line 214: The default value is `10` in the values.yaml

ywk253100 (approved):

heww (approved):

Issue #731, opened 2020-09-17
413 Payload Too Large on login (ingress)


0pendev:
When deploying ingress with the following command : 
`
    helm install skube-registry harbor/harbor\
	 --namespace skube-system\
	 --set externalURL=https://harbor.local\
	 --set expose.type=ingress\
	 --set expose.ingress.hosts.core=harbor.local\
	 --set expose.ingress.hosts.notary=harbor.local
`

I can access the login page but when trying to login the HTTP request fails with the following error : 
`
413 Payload Too Large
The request is too large. 
`

0pendev:
For anyone with this issue just use the following values : 
```yaml
    externalURL: https://harbor.local
    expose:
      type: ingress
      ingress:
        hosts:
          core: harbor.local
          notary: notary.local
        annotations:
          #Update sizes for your needs
          ingress.kubernetes.io/proxy-body-size: "1m"
          ingress.kubernetes.io/ssl-redirect: "true"
          nginx.ingress.kubernetes.io/proxy-body-size: "1m"
          nginx.ingress.kubernetes.io/ssl-redirect: "true"

```

ywk253100:
Seems that the value of `ingress.kubernetes.io/proxy-body-size` varies for different kinds of ingress controller. `0` is a valid value for nginx ingress controller.

As we have expose the `ingress.kubernetes.io/proxy-body-size` in the `values.yaml`, so you can change it when installing chart.

I'm closing this issue, feel free to reopen it

0pendev:
> Seems that the value of `ingress.kubernetes.io/proxy-body-size` varies for different kinds of ingress controller. `0` is a valid value for nginx ingress controller.
> 
> As we have expose the `ingress.kubernetes.io/proxy-body-size` in the `values.yaml`, so you can change it when installing chart.
> 
> I'm closing this issue, feel free to reopen it

It might be interesting to specify in the documentation that the ingress that should be used is NGINX, as configurations are not applicable to any ingress service.

PR #737, opened 2020-09-20
Fixes proxy-body-size for ingress deployment

0pendev:
In the default values the ingress proxy-body-size was set to 0. It makes an ingress deployment broken. The default value as define in the NGINX documentation is 1m. 0 makes no sense here.

Fixes issue #731

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -44,7 +44,7 @@ expose:
       ingress.kubernetes.io/ssl-redirect: "true"
       ingress.kubernetes.io/proxy-body-size: "1m"
       nginx.ingress.kubernetes.io/ssl-redirect: "true"
-      nginx.ingress.kubernetes.io/proxy-body-size: "1m"
+      nginx.ingress.kubernetes.io/proxy-body-size: "0"
   clusterIP:
     # The name of ClusterIP service
     name: harbor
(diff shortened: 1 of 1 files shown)


0pendev:
A  proxy-body-size of 0 seems to be required for a nginx ingress. I do not know why. My environment was using haproxy and gave an appropriate 413 error code. I've let the nginx annotation to 0 for it to work in the unit tests. Some tests my be needed for other [ingress controllers](https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/) like traefik.

Waiting for review.

whereismyjetpack:
I think `0` is the desired value here, as it should be disabling the check.

"Setting size to 0 disables checking of client request body size. "

http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

0pendev:
Isn't it a value specific to NGINX ? And shouldn't there only be the value 0 for the annotation specific to NGINX ?

It is not a standard value. For example, haproxy see this value as 0 and not a way to disable the check.

ywk253100:
@0pendev Please see https://github.com/goharbor/harbor-helm/issues/731

I'm closing this PR. Thanks anyway

PR #753, opened 2020-09-28
Nginx secret aren't required if not ingress and tls

bclermont:


bclermont:
https://github.com/goharbor/community/issues/132 I will never contribute any improvement to harbor anymore

PR #760, opened 2020-10-11
handle nginx behind another proxy

jouve:
nginx configuration does not work when it's behind another proxy as noted in the comment in nginx's configmap:
```
           # When setting up Harbor behind other proxy, such as an Nginx instance, remove the below line if the proxy already has similar settings.
          proxy_set_header X-Forwarded-Proto $scheme;
```

I'm in that case where my haproxy already defines `X-Forwarded-Proto`, so I suggest reusing `X-Forwarded-Proto` if it's already defined instead of asking people "to remove the below line" ;)

### DIFF
--- a/templates/nginx/configmap-http.yaml
+++ b/templates/nginx/configmap-http.yaml
@@ -43,6 +43,11 @@ data:
 
       access_log /dev/stdout timed_combined;
 
+      map $http_x_forwarded_proto $x_forwarded_proto {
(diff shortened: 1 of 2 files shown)


jouve:
@reasonerjt  is this change stil relevant after your PR on  goharbor/harbor#13835 ?

reasonerjt (approved):

ywk253100 (approved):

PR #762, opened 2020-10-13
Generate more values, protect them for the upgrade

jsuchome:
Allow generation of more random values where it makes sense. 
On upgrade, fetch the existing ones and do not randomize again.

Signed-off-by: Jiří Suchomel <jiri.suchomel@suse.com>

### DIFF
--- a/templates/NOTES.txt
+++ b/templates/NOTES.txt
@@ -1,3 +1,15 @@
 Please wait for several minutes for Harbor deployment to complete.
 Then you should be able to visit the Harbor portal at {{ .Values.externalURL }}
 For more details, please visit https://github.com/goharbor/harbor
+
+{{ "" }}
+{{- if .Release.IsInstall }}
+  {{- if eq "Harbor12345" .Values.harborAdminPassword }}
+The default password was set for Administrator user. Change it as soon as you log in to the Portal service.
+  {{- else if eq "" .Values.harborAdminPassword }}
+The password for Portal Administrator user was generated randomly. Fetch it by running this command:
+{{ "" }}
+kubectl get secret {{ template "harbor.core" . }} -n {{.Release.Namespace}} -o jsonpath="{.data.HARBOR_ADMIN_PASSWORD}" | base64 --decode
+{{ "" }}
+ {{- end }}
+{{- end }}
--- a/templates/core/core-secret.yaml
+++ b/templates/core/core-secret.yaml
@@ -6,7 +6,7 @@ metadata:
 {{ include "harbor.labels" . | indent 4 }}
 type: Opaque
 data:
-  secretKey: {{ .Values.secretKey | b64enc | quote }}
+  secretKey: {{ .Values.core.secretKey | default (randAlphaNum 16) | b64enc | quote }}
   secret: {{ .Values.core.secret | default (randAlphaNum 16) | b64enc | quote }}
 {{- if not .Values.core.secretName }}
   tls.crt: {{ .Files.Get "cert/tls.crt" | b64enc }}
--- a/templates/core/core-secret.yaml
+++ b/templates/core/core-secret.yaml
@@ -1,18 +1,51 @@
+{{- $secret := include "harbor.core" . }}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ template "harbor.core" . }}
+  name: {{ $secret }}
   labels:
 {{ include "harbor.labels" . | indent 4 }}
 type: Opaque
 data:
+{{- if .Release.IsUpgrade }}
+  {{- $prevSecret := (lookup "v1" "Secret" .Release.Namespace $secret) }}
+  {{- if .Values.secretKey }}
+  secretKey: {{ .Values.secretKey | b64enc | quote }}
+  {{- else if $prevSecret }}
+  secretKey: {{ $prevSecret.data.secretKey }}
+  {{- else }}
+  secretKey: {{ randAlphaNum 16 | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.secret }}
+  secret: {{ .Values.secret | b64enc | quote }}
+  {{- else if $prevSecret }}
+  secret: {{ $prevSecret.data.secret }}
+  {{- else }}
+  secret: {{ randAlphaNum 16 | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.harborAdminPassword }}
+  HARBOR_ADMIN_PASSWORD: {{ .Values.harborAdminPassword | b64enc | quote }}
+  {{- else if $prevSecret }}
+  HARBOR_ADMIN_PASSWORD: {{ $prevSecret.data.HARBOR_ADMIN_PASSWORD }}
+  {{- else }}
+  HARBOR_ADMIN_PASSWORD: {{ randAlphaNum 16 | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.core.xsrfKey }}
+  CSRF_KEY: {{ .Values.core.xsrfKey | b64enc | quote }}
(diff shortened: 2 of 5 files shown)


templates/NOTES.txt
```diff
@@ -1,3 +1,15 @@
...
 For more details, please visit https://github.com/goharbor/harbor
+
+{{ "" }}
+{{- if .Release.IsInstall }}
+  {{- if eq "Harbor12345" .Values.harborAdminPassword }}
+The default password was set for Administrator user. Change it as soon as you log in to the Portal service.
+  {{- else if eq "" .Values.harborAdminPassword }}
+The password for Portal Administrator user was generated randomly. Fetch it by running this command:
+{{ "" }}
+kubectl get secret {{ template "harbor.core" . }} -n {{.Release.Namespace}} -o jsonpath="{.data.HARBOR_ADMIN_PASSWORD}" | base64 --decode
```
stefannica, line 12: Would it make sense to print the password value itself, instead of the instructions ?

templates/core/core-secret.yaml
```diff
@@ -1,18 +1,26 @@
...
 type: Opaque
 data:
-  secretKey: {{ .Values.secretKey | b64enc | quote }}
+{{- if .Release.IsInstall }}
+  secretKey: {{ .Values.core.secretKey | default (randAlphaNum 16) | b64enc | quote }}
   secret: {{ .Values.core.secret | default (randAlphaNum 16) | b64enc | quote }}
+  HARBOR_ADMIN_PASSWORD: {{ .Values.harborAdminPassword | default (randAlphaNum 16) | b64enc | quote }}
+  CSRF_KEY: {{ .Values.core.xsrfKey | default (randAlphaNum 32) | b64enc | quote }}
+{{- else }}
+  secretKey: {{ (lookup "v1" "Secret" .Release.Namespace $secret).data.secretKey }}
```
stefannica, line 16: For some keys can passwords that _can_ be changed by the admin after the initial deployment by updating values.yaml (secretKey, secret, CSRF_KEY), I think it makes sense to use the values.yaml value even during an upgrade. This definitely doesn't apply to the harbor admin password.

templates/core/core-secret.yaml
```diff
@@ -1,18 +1,26 @@
...
+  secret: {{ (lookup "v1" "Secret" .Release.Namespace $secret).data.secret }}
+  HARBOR_ADMIN_PASSWORD: {{ (lookup "v1" "Secret" .Release.Namespace $secret).data.HARBOR_ADMIN_PASSWORD }}
```
stefannica, line 18: I found this in the [helm documentation](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/):

_Keep in mind that Helm is not supposed to contact the Kubernetes API Server during a helm template or a helm install|update|delete|rollback --dry-run, so the lookup function will return an empty list (i.e. dict) in such a case._

templates/NOTES.txt
jsuchome: I think the Notes.text is rendered before the secret is created.

templates/core/core-secret.yaml
jsuchome: So ... ? `--dry-run` is not a problem, I assume

jsuchome: Hm, but that `lookup` during template call is probably the reason why `lint` fails...

jsuchome:
2.16.8 lint fails with `function "lookup" not defined`. This should be addressed by https://github.com/helm/helm/pull/7969 and if I read it correctly, should be fixed with 2.16.9

stefannica:
A really nice benefit of re-using the password and key values in the existing k8s secrets is that a `helm upgrade` command would no longer affect services unless they actually need to be re-deployed due to other configuration changes. But in order to really rip those benefits, you'd have to do the same for the jobservice and registry secrets. Currently, some values are still generated there with every `helm upgrade`, even if no real helm chart changes are introduced:

- in the jobservice secret: 
```
  JOBSERVICE_SECRET: {{ .Values.jobservice.secret | default (randAlphaNum 16) | b64enc | quote }}
```
- and n the registry secret:
``` 
  REGISTRY_HTTP_SECRET: {{ .Values.registry.secret | default (randAlphaNum 16) | b64enc | quote }} 
```

templates/jobservice/jobservice-secrets.yaml
```diff
@@ -1,10 +1,22 @@
...
-  name: "{{ template "harbor.jobservice" . }}"
+  name: {{ $secret }}
   labels:
 {{ include "harbor.labels" . | indent 4 }}
 type: Opaque
 data:
+  {{- if .Release.IsUpgrade }}
+    {{- $prevSecret := (lookup "v1" "Secret" .Release.Namespace $secret) }}
+    {{- if .Values.jobservice.secret }}
+    JOBSERVICE_SECRET: {{ .Values.jobservice.secret | b64enc | quote }}
```
stefannica, line 13: The indentation here is incorrect. Only detectable during `helm upgrade`.

templates/jobservice/jobservice-secrets.yaml
```diff
@@ -1,10 +1,22 @@
...
+    {{- else if $prevSecret }}
+    JOBSERVICE_SECRET: {{ $prevSecret.data.JOBSERVICE_SECRET }}
```
stefannica, line 15: Same here

templates/jobservice/jobservice-secrets.yaml
```diff
@@ -1,10 +1,22 @@
...
+    {{- else }}
+    JOBSERVICE_SECRET: {{ randAlphaNum 16 | b64enc | quote }}
```
stefannica, line 17: And here

templates/registry/registry-secret.yaml
```diff
@@ -1,13 +1,25 @@
...
+  name: {{ $secret }}
   labels:
 {{ include "harbor.labels" . | indent 4 }}
 type: Opaque
 data:
-  REGISTRY_HTPASSWD: {{ .Values.registry.credentials.htpasswd | b64enc | quote }}
+  {{- if .Release.IsUpgrade }}
+    {{- $prevSecret := (lookup "v1" "Secret" .Release.Namespace $secret) }}
+    {{- if .Values.registry.secret }}
+    REGISTRY_HTTP_SECRET: {{ .Values.registry.secret | b64enc | quote }}
```
stefannica, line 13: The indentation here is incorrect. Only detectable during `helm upgrade`.

templates/registry/registry-secret.yaml
```diff
@@ -1,13 +1,25 @@
...
+    {{- else if $prevSecret }}
+    REGISTRY_HTTP_SECRET: {{ $prevSecret.data.REGISTRY_HTTP_SECRET }}
```
stefannica, line 15: Same here

templates/registry/registry-secret.yaml
```diff
@@ -1,13 +1,25 @@
...
+    {{- else }}
+    REGISTRY_HTTP_SECRET: {{ randAlphaNum 16 | b64enc | quote }}
```
stefannica, line 17: And here

stefannica (changes requested):

templates/jobservice/jobservice-secrets.yaml
jsuchome: Yeah, I did it on purpose; does it break anything?

stefannica (approved):

ywk253100:
The function `lookup` introduced in this PR is supported only by helm v2.
I'm putting it into the backlog and move it out of 1.6 scope.

We'll consider it again when we deprecate the support for helm v2

PR #764, opened 2020-10-15
Support deployment annotations for nginx

JulienPotvin:
Some controllers use deployment annotations.


reasonerjt:
Thanks for the PR but could you provide more context and justification?  What's the impact if this is not merged?

We want to refrain from adding too many attributes into values.yaml

PR #766, opened 2020-10-15
Security hardening: generate password for internal database

stefannica:
To harden the security for the internal database service, a strong
postgresql password is generated if one is not explicitly configured
in values.yaml.

This requires several helm chart changes to ensure that the
password is not re-generated with every `helm upgrade`:
* the database k8s secret is used to store not only the postgreql
password, but also all other database connection URL variables that
incorporate it. To keep the implementation simple, and at the same
time to ensure that sensitive data is stored in a secret rather
than in config-maps or as explicit environment variable values,
the database secret is also used to store the password and URL
values used for the external database case
* the template logic gives preference to the database password value
in values.yaml, if one is present. To avoid re-generating the
password during `helm upgrade`, the previous internal password value is
extracted from the existing secret, if present.
* an annotation is added to all deployments and stateful-sets that
reference the redis password or one of the connection URLs, to ensure
they are redeployed if the password changes. The fact that the
password is not re-generated with each `helm upgrade` reduces the
set of impacted workload objects that need to be updated during
an upgrade.
* the database password and URL values used for the various consumer
components (core, clair, notary) are removed from associated secrets
and explicit environment variables and replaced with `env.valueFrom`
statements pointing to the database secret data. This is needed,
because the database password (and connection URL values that include
it) can only be referenced *once* if auto-generated, because helm
lacks a notion of global static variables that can be set from within
the templates. In some cases (clair and notary) this also requires the
database URL to be injected into the configuration files using an init
container
* as already mentioned, using a named template to generate the random
password for database does not work well when the password is referenced
from multiple places, because the template generates a different password
each time it is invoked. For this reason, all named templates that use
the database password value in one way or another (e.g.
`harbor.database.notaryServer`, `harbor.database.notarySigner`)
have been modified to expect the password to be supplied as a
`.databasePassword` parameter instead, and these templates are called
from a single place: the database secret template.

Signed-off-by: Stefan Nica <snica@suse.com>


jsuchome (approved):
This looks good. The lint error is only seen in the old version of Helm and I hope it could be ignored.

PR #770, opened 2020-10-24
Fixes harbor/harbor#8224

chris-mccoy:
Changes the mode of the postgres database directory to remove group and others permissions.  This is needed on some container storage interfaces such as rook-ceph with CephFS and RADOS block device as they create directories with 0777 permissions by default, which PostgreSQL doesn't like.


chris-mccoy:
I did not sign off properly and messed up the pull request.  Apologies.  Let's try again.

PR #773, opened 2020-10-26
enable startupProbe only on k8s > 1.18

cgroschupp:
Startupprobe is activated by default only with k8s version > 1.18.0, see: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

on k8s 1.15, i got this error: `Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.containers[0]): unknown field "startupProbe" in io.k8s.api.core.v1.C`

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -42,6 +42,7 @@ spec:
       - name: core
         image: {{ .Values.core.image.repository }}:{{ .Values.core.image.tag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
+        {{- if semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion }}
         startupProbe:
           httpGet:
             path: /api/v2.0/ping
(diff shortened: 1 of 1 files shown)


cgroschupp:
@ywk253100 i think the feature-gates startupProbe is not enabled for the kind setup below k8s version 1.18. https://github.com/goharbor/harbor-helm/blob/master/test/integration/kind-cluster.yaml I'm right?

jkroepke:
> i think the feature-gates startupProbe is not enabled for the kind setup below k8s version 1.18.

See: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

Alpha State (disabled by default): 1.16, 1.17
Beta State (enabled by default): 1.18+

I implement a alternative here: 

https://github.com/goharbor/harbor-helm/pull/781

Use let the user choice if they want to use a startup probe or not.

ywk253100:
PR #781 is merged, closing this one.

@cgroschupp Thanks anyway

PR #781, opened 2020-10-30
Make startupProbe optional

jkroepke:
Alternative to: https://github.com/goharbor/harbor-helm/pull/773

Instead try to detect a correct k8s version, just decide the user if they want to use startupProbes.

startupProbes are implement as Feature Gate but disabled by default since 1.16.

Since 1.18, `startupProbes` are enabled by default.

Using a strict `if version >= 1.18` check my help in general but users they enabled startupProbes in 1.16 could not benefit from.

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -43,6 +43,7 @@ spec:
       - name: core
         image: {{ .Values.core.image.repository }}:{{ .Values.core.image.tag }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
+        {{- if .Values.core.startupProbe.enabled }}
         startupProbe:
           httpGet:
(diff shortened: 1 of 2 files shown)


reasonerjt (approved):

ywk253100:
The reason why we make `startupProbe.initialDelaySeconds` configurable is that when `core` starts up, it runs the upgrading process which may take a very long time, in this case the `core` will be restarted  because of the liveness probe failure.
So if the startup probe is disabled in some Kubernetes deployment, it isn't possible to configure the `initialDelaySeconds` any more, which means users have no method to avoid the restarting of `core`

reasonerjt:
Maybe add a comment in the values.yaml
If the user really needs to disable that, they need to consider the risk?

ywk253100 (approved):

Issue #786, opened 2020-11-06
No control for rotating auto-generated certificates


stefannica:
When auto-generated TLS certificates are used for external and internal communication, these certificates will be re-generated with every helm run, which has some undesired effects:

* external TLS certificates being re-generated during upgrades, which requires users to re-configure their clients to accept the new CA certificate. As stated in the upgrade documentation, this can be addressed by making a copy of the existing secret and providing that as a `expose.tls.secret.secretName` value, but this is one complication that could be avoided by extending the helm chart with an option to inhibit certificate rotation on upgrade.
* "helm upgrade" is also used for Day 2 operations other than Harbor upgrades, such as making targeted configuration changes (e.g. password changes for the external database or redis services, changing the resources allocated for a particular component). It is expected that such operations only restart the pods of the affected services, to reduce downtime. However, when internal TLS is enabled with auto-generated certificates, almost all internal services are restarted upon "harbor upgrade", leading to unnecessary load and downtime

The Harbor helm chart should provide an option to control when and if auto-generated certificates are rotated.

PR #787, opened 2020-11-06
Add values.yaml options to control TLS certificate rotation

stefannica:
When auto-generated TLS certificates are used for external and internal
TLS, these certificates will be re-generated with every helm run, which
can have undesired effects, such as:
* external TLS certificates being re-generated during upgrades, which
requires users to re-configure their clients to accept the new CA
certificate. This can currently be alleviated by making a copy of the
existing secret and providing that as a secret certSource, but it's
an unnecessary complication that can be avoided by extending the
helm chart templates to cover this case.
* "helm upgrade" is also used for Day 2 operations other than upgrades,
such as making targeted configuration changes (e.g. password changes
for the external database or redis services, resizing the persistent
storage volume for a particular component). It is expected that such
operations only restart the pods of the affected services, to reduce
downtime. However, when internal TLS is enabled with auto-generated
certificates, almost all internal services are restarted, leading
to unnecessary load and downtime.

This change introduces a `rotateCert` configuration option to all places
that involve auto-generating certificates:
* external TLS
* internal TLS
* notary communications (also internal TLS, but one that cannot be
disabled for technical reasons)
* core token encryption/decryption - currently using a TLS certificate
and private key that are hard-coded in the helm chart (which can be
labeled as a security risk)

The `rotateCert` configuration option may be set to `false` in the
helm chart configuration (or supplied as a helm `--set` command line
argument) to explicitly disable re-generating TLS certificates during
a "helm upgrade" run where such a change isn't required. When set to
`false', the existing TLS certificates are loaded from their respective
secrets and reused.

Fixes: #786

Signed-off-by: Stefan Nica <snica@suse.com>


templates/core/core-secret.yaml
```diff
@@ -1,3 +1,4 @@
+{{- $secret := include "harbor.core" . }}
```
jsuchome, line 1: As you define the name here, shouldn't you use it wherever possible? (`metadata:` -> `name:` here and I assume in all other secrets...

templates/ingress/secret.yaml
```diff
@@ -1,15 +1,15 @@
 {{- if eq (include "harbor.autoGenCertForIngress" .) "true" }}
-{{- $ca := genCA "harbor-ca" 365 }}
-{{- $cert := genSignedCert .Values.expose.ingress.hosts.core nil (list .Values.expose.ingress.hosts.core .Values.expose.ingress.hosts.notary) 365 $ca }}
+  {{- $secret := include "harbor.ingress" . }}
+  {{- $ca := genCA "harbor-ca" 365 }}
+  {{- $cert := genSignedCert .Values.expose.ingress.hosts.core nil (list .Values.expose.ingress.hosts.core .Values.expose.ingress.hosts.notary) 365 $ca }}
+  {{- $certCtx := (merge (dict "ca" $ca "cert" $cert "secret" $secret "rotateCert" .Values.expose.tls.auto.rotateCert ) .) }}
```
jsuchome, line 5: Is the change of indentation in this block correct?

templates/core/core-secret.yaml
stefannica: Done

templates/ingress/secret.yaml
stefannica: Yes. Indentation doesn't really matter if you're using the `{{-` or `-}}` markers to remove whitespaces before and after the block.

jsuchome (commented):
I love the idea, and after running my own test it appears to work as described.

reasonerjt:
I have some different thoughts, currently we already support referencing a secret for the certificate that are generated by external tools.

I hope we can delegate the certificate management work to external tools from the chart.

stefannica:
> I have some different thoughts, currently we already support referencing a secret for the certificate that are generated by external tools.
> 
> I hope we can delegate the certificate management work to external tools from the chart.

@reasonerjt you're correct, the helm chart already has the option of providing custom certificates via secrets. However, this PR is  concerned only with the convenience of using self-signed generated certificates. It makes it easier for Harbor operators to make configuration changes and run upgrades with `helm upgrade` without incurring a re-generation of certificates. This is basically a huge improvement for those who don't want to go through the hassle of providing custom certificates (and we all know how painful that can get): they won't have to re-configure all registry clients every time they run `helm upgrade` because their certificates are re-generated. One place where this improvement is immediately effective is in CI/CD jobs that test Day 2 operations for Harbor, such as upgrades or changing something in the helm configuration.

chulkilee:
Can all components in harbor detect the changes of k8s Secrets when changed by other tools? If so, what about replacing own cert generation and rotation with cert-manager integration?

PR #808, opened 2020-12-09
Define registry host

jkroepke:
Allows to overrides registry host. 

Required if X-Forwarded-For or Forwarded header incorrectly implemented on infrastructure side.


jkroepke:
Could someone verify it?

PR #823, opened 2021-01-06
Fix for networking.k8s.io/v1 Ingress

chulkilee:
v1 Ingress requires changes in `rules` - so helm against k8s (e.g. 1.19.3 in my case) is failing.

This PR fixes it without updating integration test github action to test against k8s 1.19 as well.

Follow-up of #814.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -9,22 +9,22 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        k8s_version: [v1.18.2, v1.17.5, v1.16.9]
+        k8s_version: [v1.20.0, v1.19.4, v1.18.8, v1.17.11, v1.16.15]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
       - name: Setup Docker
-        uses: docker-practice/actions-setup-docker@0.0.1
+        uses: docker-practice/actions-setup-docker@v1
         with:
-          docker_version: 18.09
+          docker_version: "20.10"
           docker_channel: stable
           docker_daemon_json: '{"insecure-registries":["0.0.0.0/0"]}'
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.0.0-rc.1
+        uses: helm/kind-action@v1.1.0
         with:
-          version: v0.8.1
+          version: v0.9.0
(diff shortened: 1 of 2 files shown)


reasonerjt:
@chulkilee thanks for this PR!
We are also making adjustments wrt what k8s versions will be tested.

We want to test 3 versions, once this change is made you may need to rebase this PR.

templates/ingress/ingress.yaml
```diff
@@ -127,8 +172,15 @@ spec:
       paths:
       - path: {{ .notary_path }}
         backend:
+{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
```
ywk253100, line 175: The `pathType` should be specified when version >=1.19

.github/workflows/integration.yaml
```diff
@@ -9,22 +9,22 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        k8s_version: [v1.18.2, v1.17.5, v1.16.9]
+        k8s_version: [v1.20.0, v1.19.4, v1.18.8, v1.17.11, v1.16.15]
```
ywk253100, line 12: We want to test against only the three latest versions of k8s, so keep the `v1.20.0`, `v1.19.4` and `v1.18.8`

ywk253100:
@chulkilee Could you fix the issues according to the comments?

.github/workflows/integration.yaml
chulkilee: README says "Kubernetes cluster 1.16+" :wink:

I think it would be better to have another issue/PR for dropping 1.17 and 1.16.

templates/ingress/ingress.yaml
chulkilee: Thanks for the catch! I had tested without notary 🤦  Fixed it

ywk253100:
@chulkilee We have merged a fix for the CI failure, could you rebase your PR?

ywk253100 (approved):

.github/workflows/integration.yaml
reasonerjt: sure

reasonerjt (approved):

PR #832, opened 2021-01-12
Update the kubernetes version used in CI

ywk253100:
Update the kubernetes version used in CI

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


heww (approved):

reasonerjt (approved):

reasonerjt:
Approved, but I think we should also update the readme.md

ywk253100:
updated

Hokwang:
@reasonerjt Hi, can you tell me why kubernetes 1.18+ needs for harbor 2.2.0+ ?

reasonerjt:
We plan to only support the 3 most recently released versions of k8s.  That said, the chart may work on lower versions, but we will not verify it in the CI.

Issue #864, opened 2021-02-25
Cannot connect to registry behind HTTP proxy


carolili:
When installing the chart version 1.5.3 from the goharbor Helm repo the HTTP proxy does not work, e.g. when trying to add a new registry endpoint and test the connection. Also doing a `curl` on the `harbor-core` pod fails.

It appears that this Harbor Helm chart is defining the `*_PROXY` environment variables that are used in several components in uppercase. However, when defining the `*_PROXY` environment variables in lowercase as e.g. `http_proxy` the proxy works. Both in the CLI and when trying to add a registry and pull some images. It appears that [there is no standard way](https://unix.stackexchange.com/questions/212894/whats-the-right-format-for-the-http-proxy-environment-variable-caps-or-no-ca) to define the proxy environment variables.

I tested this with a modified helm chart that I will attach as pull request. Please let me know if it can be merged or if there is something I can change on my side.

Simple example:
```
harbor [ /harbor ]$ env | grep HTTP_PROXY
HTTP_PROXY=http://my_company_proxy.com
harbor [ /harbor ]$ curl google.com
curl: (7) Failed to connect to google.com port 80: Connection refused
harbor [ /harbor ]$ export http_proxy=http://my_company_proxy.com
harbor [ /harbor ]$ curl google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
```

alxbse:
I'm successfully using harbor behind a proxy using the proxy settings built in to this chart: https://github.com/goharbor/harbor-helm/blob/master/values.yaml#L349.

Using it to fetch images from dockerhub, quay and gcr all through proxy.

carolili:
@alxbse Thank you for the comment. Those settings do not work for me when the environment variables are uppercase. Would it be possible that there is something else in your environment that sets the lowercase version of the proxy variables?

Could you say what version of Kubernetes you are using and what container runtime + version you are using? I have the following:
```
Kubernetes: 1.19.7
Docker: 19.03
```

Perhaps it could be an issue stemming from Docker as container runtime?

alxbse:
This is an interesting issue. Exec;ed into one of my core pods to verify and there are no lower case environment variables.

```console
$ export | grep -i proxy
[...]

PR #865, opened 2021-02-25
Add lowercase proxy environment variables

carolili:
* Add lowercase proxy environment variables (`*_proxy`) to core, chartmuseum, exporter, jobservice, registry, and trivy.
* Relates to #864.


reasonerjt:
I don't think this is necessary as commented in #864

PR #872, opened 2021-02-28
(fix): expose metrics ports

lukasmrtvy:
- ServiceMonitor can not work without explicitly exposed ports on deployments. 
- Now its possible to use component label selector

```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: harbor
  labels:
    release: prometheus-stack
spec:
  selector:
    matchLabels:
      app: harbor
  endpoints:
  - targetPort: 8001
```


ninjadq:
It can collect metrics if you set port to `metric`. Reopen if you have any other ideas

PR #873, opened 2021-03-01
Update ingress rules

ywk253100:
Portal introduces a new file called "common.js", it is matched by the ingress rule of core "/c" which is mistake. This commit adds the suffix "/" for each rules to avoid such cases


heww (approved):

reasonerjt (approved):

mxmorin:
Can you provide a fixed release because we can't use UI due to this bug
Thanks in dvance

viceice:
Can we have a new release, as i also can't access the ui after upgrade to v2.3.0 images 😕

PR #882, opened 2021-03-05
fix "harbor.internalTLS.exporter.secretName" not defined

jsaulalonso:


reasonerjt:
We may not want internal TLS for exporter.

@ninjadq  thoughts?

natitomattis:
I'm having the same issue enabling TLS in the new chart version. Thanks @jsaulalonso

ninjadq:
fixed in https://github.com/goharbor/harbor-helm/pull/949/commits/7c145171e2465df3dc94a46415d7fcd72a916f72

PR #883, opened 2021-03-05
Rename service ports for compatibility with istio.

### LINKED ISSUE #77: Add support for Istio
Would be nice to add support for Istio VirtualService.

j-zimnowoda:
If sevice port names are used, Istio expects the following convention: <protocol>[-<suffix>]
If the port names does not follow the convention, then istio is not able to route traffic properly.

Fixes #77

Signed-off-by: Jehoszafat Zimnowoda <jehoszafat.zimnowoda@redkubes.com>

### DIFF
--- a/templates/core/core-svc.yaml
+++ b/templates/core/core-svc.yaml
@@ -9,11 +9,11 @@ spec:
   type: NodePort
 {{- end }}
   ports:
-    - name: web
+    - name: {{ ternary "https-web" "http-web" .Values.internalTLS.enabled }}
       port: {{ template "harbor.core.servicePort" . }}
       targetPort: {{ template "harbor.core.containerPort" . }}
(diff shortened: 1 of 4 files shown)


reasonerjt:
thanks @j-zimnowoda 
Supporting istio is out of our radar but the code change seems no negative effects.

Could you explain the usage of ternary?  Any reference?

Please also fix the CI failure.

j-zimnowoda:
@reasonerjt 

> The ternary function takes two values, and a test value. If the test value is true, the first value will be returned.
See more: http://masterminds.github.io/sprig/defaults.html

Are there special commands to trigger testing pipeline or just a new commit can do it?

reasonerjt (approved):

ywk253100 (approved):

PR #893, opened 2021-03-10
Add another init container for database to ensure the permission of data directory

### LINKED ISSUE #876: [Bug] Database data directory permission
If you are reporting a problem, please make sure the following information are provided:

**Expected behavior and actual behavior:**
After upgrading I'm getting the following error of the database instance:

```
FATAL: data directory "/var/lib/postgresql/data/pgdata" has group or world access
DETAIL: Permissions should be u=rwx (0700).
```
Perhaps it would be easy to chmod it before starting up to ensure we have the right permissions?

**Versions:**
Please specify the versions of following systems.

- harbor version: 2.1.3 -> 2.2.0
- Kubernetes: 1.19.7

**Additional context:**

The database instance uses Persisted storage backed by Ceph.

Relevant helm value:
```
    database:
      existingClaim: ""
      storageClass: "rook-ceph-block"
      subPath: "data"
      accessMode: ReadWriteOnce
      size: 1Gi
```

After debugging the issue a bit further I see in the init container it creates the data directory if it doesn't exist and ensures it's created with `0700` permissions.
Unfortunately we don't enforce it once the directory exists.
My suggestion is to replace the `|| true` at the end and convert that into `|| chmod 700 $datadirectory && true`

Note: I first reported this under https://github.com/goharbor/harbor/issues/14354 but was requested to move it here as it's specific to the helm chart deployment.

ywk253100:
Fixes #876

### DIFF
--- a/templates/database/database-ss.yaml
+++ b/templates/database/database-ss.yaml
@@ -50,6 +50,19 @@ spec:
           - name: database-data
             mountPath: /var/lib/postgresql/data
             subPath: {{ $database.subPath }}
+      # with "fsGroup" set, each time a volume is mounted, Kubernetes must recursively chown() and chmod() all the files and directories inside the volume
+      # this causes the postgresql reports the "data directory /var/lib/postgresql/data/pgdata has group or world access" issue when using some CSIs e.g. Ceph
+      # use this init container to correct the permission
+      # as "fsGroup" applied before the init container running, the container has enough permission to execute the command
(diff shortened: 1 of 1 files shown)


reasonerjt (approved):

heww (approved):

mumrau:
Maybe we should make this kind of "feature"  tweakable (i.e, optional, and opt-in based) from the `values.yaml`?

Issue #900, opened 2021-03-15
Support a subject alternate name on ingresses


ryan-a-baker:
We have a globally replicated Harbor instance, and I'd love to front it all with DNS leveraging Geo-location.

Unfortunately, I'm unable to do this because the value for the ingress host is [only a single entry](https://github.com/goharbor/harbor-helm/blob/master/templates/ingress/ingress.yaml#L60).  The helm chart should support multiple hosts so that dynamically generated certificates (such as Let's Encrypt) can have SANs.

reasonerjt:
@ryan-a-baker 
Per my understanding this is a highly customized deployment approach, I'm reluctant to merge it into master.  
I suggest that we wait until there are enough use cases like yours before we merge it.

ryan-a-baker:
I understand.  This was something I did a while back and just now opening the PR for it.  We can work around it for now.  In the meantime, I'll add more testing results and context to the issue and the PR, as I did this all in a bit of a hurry.

mothershipper:
I'd like to add a plus one to this -- we also have a globally replicated harbor deployment

sspreitzer:
Adding my 👍 to this request. Seems trivial.

cfiehe:
We are facing the same limitation. We have a new harbor deployment with a new hostname, but we have to be backward-compatible with the old hostname. The plan is that the old hostname becomes CNAME for the new one. I have modified the ingress resource, so that `ingress.hosts.core` becomes a list instead of a single hostname:
```
{{- if eq .Values.expose.type "ingress" }}
{{- $ingress := .Values.expose.ingress -}}
{{- $tls := .Values.expose.tls -}}
{{- if eq .Values.expose.ingress.controller "gce" }}
  {{- $_ := set . "portal_path" "/*" -}}
  {{- $_ := set . "api_path" "/api/*" -}}
  {{- $_ := set . "service_path" "/service/*" -}}
  {{- $_ := set . "v2_path" "/v2/*" -}}
  {{- $_ := set . "chartrepo_path" "/chartrepo/*" -}}
  {{- $_ := set . "controller_path" "/c/*" -}}
{{- else if eq .Values.expose.ingress.controller "ncp" }}
  {{- $_ := set . "portal_path" "/.*" -}}
  {{- $_ := set . "api_path" "/api/.*" -}}
  {{- $_ := set . "service_path" "/service/.*" -}}
  {{- $_ := set . "v2_path" "/v2/.*" -}}
[...]

PR #901, opened 2021-03-15
Add Subject Alternate Names for Ingress Certificates

ryan-a-baker:
This will resolve https://github.com/goharbor/harbor-helm/issues/900.  I hacked around on this a bit and got it to work, but I'm sure there are other (better) ways to handle this.  Feel free to steer me in a different direction if there is a better one.

I left the existing $ingress.hosts.core value to maintain backwards compatibility.  Was tested with and without alternate host names defined.


reasonerjt:
See my comment in #900, I hope we hold this for a while as I'm not sure this is very common usage.

PR #908, opened 2021-03-20
Disable update strategy when there is no pvc in chartmuseum and registry

WesternStar:
This fixes a bug that when using s3, etc to store images and charts the helm chart still sets a strategy for the helm chart when it isn't nessecary.


ywk253100:
@WesternStar What's the bug if the update strategy is enabled for them?

WesternStar:
Well if the images are stored in S3 and we still need the recreate strategy
for the job service this causes downtime that's unnecessary. Harbor
downtime is pretty visible for us and it would be good to avoid that.
Besides that the only reason to use recreate is if your filesystem isn't
ReadWriteMany  afaict  and we aren't using it.

Issue #441, opened 2019-11-17
API api/systeminfo/volumes does't work


qinhang3:
I deploy Harbor in Kubernetes by helm.
Then I found volumes infomation can't show in page /harbor/projects.
I found API api/systeminfo/volumes returns
`{
  "storage": {
    "total": 0,
    "free": 0
  }
}
`
.
The core module logs:
```
2019-11-17T14:26:29Z [INFO] [/replication/registry/healthcheck.go:60]: Start regular health check for registries with interval 5m0s
2019-11-17T14:26:29Z [DEBUG] [/core/filter/sessionchecker.go:22]: Mark the request as no-session
2019-11-17T14:26:29Z [DEBUG] [/core/filter/security.go:227]: OIDC CLI modifier only handles request by docker CLI or helm CLI
2019-11-17T14:26:29Z [DEBUG] [/core/filter/security.go:445]: got user information from session
2019-11-17T14:26:29Z [DEBUG] [/core/filter/security.go:451]: using local database project manager
2019-11-17T14:26:29Z [DEBUG] [/core/filter/security.go:453]: creating local database security context...
2019-11-17T14:26:29Z [WARNING] [/core/systeminfo/imagestorage/filesystem/driver.go:51]: The path /data is not found, will return zero value of capacity
2019/11/17 14:26:29 [D] [server.go:2774] | 192.168.50.139| 200 |     2.7774ms|   match| GET      /api/systeminfo/volumes   r:/api/systeminfo/volumes
```
But there is no any volumes mount at `/data` in core's pod.
By the way, I try to mount sth at /data. Then the volume info can be show on the page .
Here is my deploy script:
```
helm install --name my-release --set \
expose.type=ingress,\
expose.ingress.hosts.core=reg.xxx.com,\
persistence.resourcePolicy=,\
persistence.persistentVolumeClaim.database.existingClaim=harbor-database-pvc,\
persistence.persistentVolumeClaim.registry.existingClaim=harbor-registry-pvc2,\
persistence.persistentVolumeClaim.chartmuseum.existingClaim=harbor-chartmuseum-pvc,\
persistence.persistentVolumeClaim.jobservice.existingClaim=harbor-jobservice-pvc,\
persistence.persistentVolumeClaim.redis.existingClaim=harbor-redis-pvc,\
externalURL=https://reg.xxx.com:31817,\
expose.tls.secretName=xxx-ssl \
harbor/harbor
```
Looking forward to your reply.

[...]

PR #558, opened 2020-04-09
Add extra configuration parameters

usualstuff:
This adds the following parameters to all deployments\stateful sets:
- extraInitContainers
- extraContainers
- extraVolumeMounts
- extraVolumes

This will allow the users to extend functionality of the chart when needed.
It will allow to solve many current issues such as:
- adding custom CA certificates (e.g. `full-strict` connection for postgres)
- adding custom sidecar containers (e.g enabling TLS encryption for redis)
- adding dependency health checks to prevent multiple restarts of pods

At quick glance there is some need for this right now:
https://github.com/goharbor/harbor-helm/issues/549
https://github.com/goharbor/harbor-helm/issues/529
https://github.com/goharbor/harbor-helm/issues/445
https://github.com/goharbor/harbor-helm/pull/555
https://github.com/goharbor/harbor-helm/pull/530
https://github.com/goharbor/harbor-helm/pull/273

PR #911, opened 2021-03-24
allow extraVolumes and extraVolumeMounts on core

mcmartins:
We've deployed harbor using helm 3 on a 1.18 Kubernetes cluster:

```
NAME  	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART       	APP VERSION
harbor	harbor   	7       	2021-03-24 10:36:22.909063483 +0000 UTC	deployed	harbor-1.6.0	2.2.0      
```
Our storage for the registry, chartmuseum and jobservice is a NFS (ReadWriteMany) solution.

When we access the GUI, the storage gauge on the right appears with a NaN GB.

When using a ReadWriteMany storage solution, I believe this solves this issue: API api/systeminfo/volumes does't work #441 

By mounting the same storage as extraVolume on the core component:

```
core:
  image:
    repository: goharbor/harbor-core
    tag: v2.2.0
    ...
    # Extra volumes to add to the deployment
    extraVolumes:
      - name: registry-data
        persistentVolumeClaim:
          claimName: nfs-harbor-dev-pvc
    # Extra volume mounts to add to the container
    extraVolumeMounts:
      - mountPath: /data
        name: registry-data
        subPath: core
```

Thanks,
Manuel

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -124,6 +124,9 @@ spec:
         {{- if .Values.caBundleSecretName }}
 {{ include "harbor.caBundleVolumeMount" . | indent 8 }}
         {{- end }}
+{{- with .Values.core.extraVolumeMounts }}
+{{ tpl (. | toYaml) $ | indent 8 }}
+{{- end }}
 {{- if .Values.core.resources }}
         resources:
 {{ toYaml .Values.core.resources | indent 10 }}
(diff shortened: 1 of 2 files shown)


mcmartins:
Just looking around now and I believe the Pull Request `Add extra configuration parameters #558` includes this one, so I'm happy if this one is ignored and the #558 is progressed ;)

rgarcia89:
Where are we with this MR?
I could also need this to mount secrets via the Hashicorp Vault CSI into the core.

Kajot-dev:
We could allow this also on portal containers since there is "setting.json" inside which could be modified that way

Jacq:
This looked promising but was never merged nor rejected, this approach could be apply for all components, for trivy it has been just merged in  #2350.

PR #914, opened 2021-03-29
Add support for ALB ingress controller

Puneeth-n:


skaymakca:
I was about to open an issue to start talking about AWS ALB ingresses and saw this PR.  I independently got this working on my own EKS cluster and think you need to change the couple of service definitions that check for gce to use NodePorts.  My ALB setup required the same ones to be NodePorts (due to ALB's requirement of NodePorts).

Puneeth-n:
which version of alb load balancer controller do you have? this seems to work fine on mine

skaymakca:
I am running:

`Image:         amazon/aws-alb-ingress-controller:v2.1.3`

I checked out your commit via git and installed the chart while setting the ingress controller to `alb`. The ingress did not reconcile/activate.  I am intending on running the ingress with the following annotations:

      kubernetes.io/ingress.class: alb
      alb.ingress.kubernetes.io/scheme: internal
      alb.ingress.kubernetes.io/certificate-arn: arn:XXXX

Running this the ingress controller had the following error:

```
Events:
  Type     Reason             Age                From     Message
  ----     ------             ----               ----     -------
  Warning  FailedDeployModel  8s (x13 over 36s)  ingress  Failed deploy model due to InvalidParameter: 1 validation error(s) found.
- minimum field value of 1, CreateTargetGroupInput.Port.
```

Due to the lack of NodePort services.  Rerunning as an ingress controller of `gce` worked with the ingress activating but didn't have the fix of moving the portal path definition being last, as your commit does for the `alb` controller option.

Alternatively adding the annotation to switch the alb target to the pods, seems to have the ingress activate.  The core pod's health checks don't pass / traffic doesn't route, but that could be an issue with security groups on my end:

    alb.ingress.kubernetes.io/target-type: ip

Puneeth-n:
@skaymakca 

These annotations work perfectly fine for me.


```hcl
annotations = {
  "kubernetes.io/ingress.class"               = "alb"
  "alb.ingress.kubernetes.io/scheme"          = "internal"
  "alb.ingress.kubernetes.io/listen-ports"    = jsonencode([{ HTTPS = 443 }])
  "alb.ingress.kubernetes.io/group.name"      = "infra"
  "alb.ingress.kubernetes.io/certificate-arn" = aws_acm_certificate.subdomain.arn
  "alb.ingress.kubernetes.io/target-type"     = "ip"
  "alb.ingress.kubernetes.io/success-codes"   = "200-499"
}
```

templates/ingress/ingress.yaml
```diff
@@ -64,10 +64,12 @@ spec:
   - http:
       paths:
 {{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
+      {{- if ne .Values.expose.ingress.controller "alb" }}
```
ywk253100, line 67: What's the difference between the `if ne .Values.expose.ingress.controller "alb"` and `eq .Values.expose.ingress.controller "alb"` code block?

Puneeth-n: @ywk253100 The important thing is where portal_path `/*` is set in the ingress configuration for alb-load-balancer-controller. if it is the first thing, then all requests go to portal_path even though other path patterns such as `api/*` exists. To avoid this, when it is an alb load balancer, I set it as the last rule.

I do not know how the other ingress work and didn't want to break them. Therefore, when ingress type is not `alb`, portal_path is the 1st rule (like before) and if it is `alb`, it is the last rule

skaymakca:
> @skaymakca
> 
> These annotations work perfectly fine for me.
> 
> ```terraform
> annotations = {
>   "kubernetes.io/ingress.class"               = "alb"
>   "alb.ingress.kubernetes.io/scheme"          = "internal"
>   "alb.ingress.kubernetes.io/listen-ports"    = jsonencode([{ HTTPS = 443 }])
>   "alb.ingress.kubernetes.io/group.name"      = "infra"
>   "alb.ingress.kubernetes.io/certificate-arn" = aws_acm_certificate.subdomain.arn
>   "alb.ingress.kubernetes.io/target-type"     = "ip"
>   "alb.ingress.kubernetes.io/success-codes"   = "200-499"
> }
> ```

Thanks for sharing your configuration, it worked on my EKS cluster.

templates/ingress/ingress.yaml
ywk253100: Per my understanding, as the portal path `/*` can match all requests, if it works as the first rule, it should work as the last one.
So how about putting the portal path as the last one for all kinds of ingress controller?

shinji62:
Can we get this merge ? using alb controller too here.

templates/ingress/ingress.yaml
Puneeth-n: @ywk253100 sure. I will update the PR

ywk253100 (approved):

jgregmac:
I appreciate seeing people work on official support the the AWS Load Balance Controller, as I am struggling with getting Harbor to work with this Ingress, too.

However, I will note that this PR WILL NOT work correctly with the AWS Load Balancer Controller release 2.2.  The reason is that the 2.2 release introduces correct support for the `pathType` value of an Ingress path.  Previously, the AWS Ingress Controller ignored this field.  Under v2.2 if you set `pathType: Prefix` and then attempt to set a path with a wildcard in it, the controller will throw errors when attempting to reconcile the manifest.

The configuration in this PR only will word with AWS Load Balancer Controller v2.1.3 and earlier.

reasonerjt:
@Puneeth-n do you have bandwidth to incoporate the change suggested in this comment:
https://github.com/goharbor/harbor-helm/pull/914#issuecomment-841841184

Puneeth-n:
@reasonerjt sure

reasonerjt:
Thanks, but I think the `pathType` should be `ImplementationSpecific` only when it's using `alb`?  I'm not sure whether other ingress controllers can handle `ImplementationSpecific` correctly, sorry we don't have the bandwidth to verify all ingress controllers, so we better minimize the scope of the change.

You may use something like a ternary to make sure `ImplementationSpecific` is applied only when the ingress controller is alb.

ywk253100:
> I appreciate seeing people work on official support the the AWS Load Balance Controller, as I am struggling with getting Harbor to work with this Ingress, too.
> 
> However, I will note that this PR WILL NOT work correctly with the AWS Load Balancer Controller release 2.2. The reason is that the 2.2 release introduces correct support for the `pathType` value of an Ingress path. Previously, the AWS Ingress Controller ignored this field. Under v2.2 if you set `pathType: Prefix` and then attempt to set a path with a wildcard in it, the controller will throw errors when attempting to reconcile the manifest.
> 
> The configuration in this PR only will work with AWS Load Balancer Controller v2.1.3 and earlier. Post v2.2.0, the default Ingress syntax is valid, although as observed elsewhere the root path matching rule "/" for the "portal" pod needs to be moved to be bottom of the paths list, or it takes priority over all other paths in the Ingress configuration.
> 
> A configuration that will work for _both_ v2.1.3- _and_ v2.2.0+ would be to use the wildcard paths, and set the `pathType: ImplementationSpecific`

@jgregmac Is the change in aws ingress controller a bug or breaking change? If it is, fixing it in the aws ingress controller makes more sense?

reasonerjt:
Removing it from 1.7.0 release, due to unresolved comment, we'll revisit this in future release.

democracytoday:
You will also need to annotate the `harbor-core` service to correctly configure the health check as follows:
```
apiVersion: v1
kind: Service
metadata:
  name: harbor-core
  annotations:
    alb.ingress.kubernetes.io/healthcheck-path: /v2
    alb.ingress.kubernetes.io/success-codes: "401"
```

jgregmac:
> > I appreciate seeing people work on official support the the AWS Load Balance Controller, as I am struggling with getting Harbor to work with this Ingress, too.
> > However, I will note that this PR WILL NOT work correctly with the AWS Load Balancer Controller release 2.2. The reason is that the 2.2 release introduces correct support for the `pathType` value of an Ingress path. Previously, the AWS Ingress Controller ignored this field. Under v2.2 if you set `pathType: Prefix` and then attempt to set a path with a wildcard in it, the controller will throw errors when attempting to reconcile the manifest.
> > The configuration in this PR only will work with AWS Load Balancer Controller v2.1.3 and earlier. Post v2.2.0, the default Ingress syntax is valid, although as observed elsewhere the root path matching rule "/" for the "portal" pod needs to be moved to be bottom of the paths list, or it takes priority over all other paths in the Ingress configuration.
> > A configuration that will work for _both_ v2.1.3- _and_ v2.2.0+ would be to use the wildcard paths, and set the `pathType: ImplementationSpecific`
> 
> @jgregmac Is the change in aws ingress controller a bug or breaking change? If it is, fixing it in the aws ingress controller makes more sense?

Sorry for the delay in reply: I have not been tracking this closely. 

I would not describe the AWS change as a bug. They have corrected erroneous behavior in their product, and have implemented canonically correct behavior for ‘pathType: prefix’. It does not seem like a breaking change either. If a bug fix breaks someone’s workaround, we don’t usually call that a “breaking change”.

zyyw:
closing this because of the merge of this following PR:
* https://github.com/goharbor/harbor-helm/pull/1206

PR #923, opened 2021-04-05
Allow setting internalTLS settings for exporter

### LINKED ISSUE #909: internalTLS broken in 1.6.0 due to missing exporter template
The new exporter component for 1.6.0 breaks the use of `internalTLS` because of a reference to a template (`harbor.internalTLS.exporter.secretName`) that does not exist:

https://github.com/goharbor/harbor-helm/blob/v1.6.0/templates/exporter/exporter-dpl.yaml#L74

It looks like some copy/paste was missing?

poblahblahblah:
Fixes #909

### DIFF
--- a/templates/exporter/exporter-dpl.yaml
+++ b/templates/exporter/exporter-dpl.yaml
@@ -68,11 +68,6 @@ spec:
       - name: config
         secret:
           secretName: "{{ template "harbor.exporter" . }}"
-      {{- if .Values.internalTLS.enabled }}
-      - name: exporter-internal-certs
-        secret:
-          secretName: {{ template "harbor.internalTLS.exporter.secretName" . }}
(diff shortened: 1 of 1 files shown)


poblahblahblah:
Something to note is that I couldn't figure out if TLS was actually used for the exporter. I suspect it's not, in which case the deployment template should be updated to not look for the key. I can do that if that's desired.

heww:
I think we can't remove the root ca of the internal tls in the exporter because it will access the core API to fetch some info.

ninjadq (approved):
lgtm

ninjadq:
fixed in this commit https://github.com/goharbor/harbor-helm/pull/949/commits/7c145171e2465df3dc94a46415d7fcd72a916f72

Issue #928, opened 2021-04-13
Database initContainers need options for resource limits


djtaylor:
In our Kube cluster, we set resource quotas at the namespace level for harbor. Each initContainer in the database statefulset needs to have the option to define resource requests/limits or else the statefulset will throw the error: `failed quota: compute-resources: must specify limits.cpu,limits.memory,requests.cpu,requests.memory`. Manually editing the statefulset works around the issues and the pod then starts.

djtaylor:
Proposed fix: https://github.com/goharbor/harbor-helm/pull/929

PR #929, opened 2021-04-13
Optional resources for db init containers

djtaylor:
This adds the option to specify resource definitions for both the migrator and permissions initContainers for the database statefulset. In a cluster where there are resource quotas set at the namespace level, these containers will fail to start. When describing the statefulset, it will complain that these resource requests/limits are required.

Fixes: https://github.com/goharbor/harbor-helm/issues/928


values.yaml
```diff
@@ -654,6 +654,17 @@ database:
     nodeSelector: {}
     tolerations: []
     affinity: {}
+    initContainer:
```
reasonerjt, line 657: how'bout comment the whole section of `initContainers`

reasonerjt (approved):

ywk253100 (approved):

PR #932, opened 2021-04-19
Add ipv6 support for nginx

ninjadq:


heww (approved):

ywk253100 (approved):

james-d-elliott:
Just a note that this change seems to break on nodes without ipv6 enabled. Would it be worth adding a disable ipv6 option?

PR #936, opened 2021-04-27
add custom annotation support for harbor and notary ingresses

### LINKED ISSUE #915: Allow Notary Ingress to Have Separate Set of Annotations
Hi,

I'd like to contribute to Harbor helm chart by allowing Notary ingress to have extra set of annotations.

For one of our requirements, we need to restrict Notary server access to a certain set of IP blocks. Currently, we'll have to do this manually by patching the ingress which is not the best long-term solution. 

My proposal would be to add the following:
`values.yaml`:
```
ingress:
  ...
  notary:
    annotations:
```

and for the template, something like:
`templates/ingress.yaml`:
```
kind: Ingress
metadata:
  name: "{{ template "harbor.ingress-notary" . }}"
  labels:
  annotations:
{{ toYaml $ingress.annotations | indent 4 }}
{{ toYaml $ingress.notary.annotations | indent 4 }}
{{- if eq .Values.expose.ingress.controller "ncp" }}
...
```

Same structure could be applied to habor ingress as well.

Let me know if any questions/concerns.
Thank you

neocorp:
Hi,
This PR includes the changes to helm chart for allowing custom annotations separately for harbor and notary ingresses as opened with the issue #915 

Following structure is the way to use it:
```
ingress:
  ...
  annotations: # common annotations for ingresses
  harbor:
    annotations: # custom annotations for harbor ingress
  notary:
    annotations: # custom annotations for notary ingress
```

Closes #915

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,9 @@ The following table lists the configurable parameters of the Harbor chart and th
 | `expose.ingress.hosts.core` | The host of Harbor core service in ingress rule | `core.harbor.domain` |
 | `expose.ingress.hosts.notary` | The host of Harbor Notary service in ingress rule | `notary.harbor.domain` |
 | `expose.ingress.controller` | The ingress controller type. Currently supports `default`, `gce` and `ncp` | `default` |
-| `expose.ingress.annotations` | The annotations used in ingress |  |
+| `expose.ingress.annotations` | The annotations used commonly for ingresses |  |
+| `expose.ingress.harbor.annotations` | The annotations specific to harbor ingress | {} |
+| `expose.ingress.notary.annotations` | The annotations specific to notary ingress | {} |
 | `expose.clusterIP.name` | The name of ClusterIP service | `harbor` |
 | `expose.clusterIP.annotations` | The annotations attached to the ClusterIP service | {} |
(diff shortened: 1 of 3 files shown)


Vad1mo (approved):

ywk253100 (approved):

Vad1mo:
The reason is to specify mainly unique set of annotations per ingress to manage these ingresses differently.

For example:
- Public access to harbor but restricted access to notary within a certain IP address range, like using nginx.ingress.kubernetes.io/whitelist-source-range: annotation

Currently, we manually patch this ingress to get it to working but having the ability to specify the annotation while installing is of course preferred.

The currently available global annotation setting applies annotations across all ingresses, however this is not always desired, so this patch adds the ability to also provide annotations on each individual ingress differently.

reasonerjt (approved):

reasonerjt (changes requested):
You need to update the values.yaml to pass the CI.

neocorp:
@reasonerjt added if statements around the annotations in template to handle when non-existent; I've already provided and empty dict `{}` in values.yaml which should be enough. I checked the tests but didn't see any regarding testing annotations. Let me know if there's anything else missing.

reasonerjt (approved):

reasonerjt:
Thanks!

PR #949, opened 2021-05-17
Fix exporter can not start issue

ninjadq:
1. cert issue when internal TLS enabled
2. add config for job service monitoring

### DIFF
--- a/templates/exporter/exporter-cm-env.yaml
+++ b/templates/exporter/exporter-cm-env.yaml
@@ -11,6 +11,7 @@ data:
   HTTPS_PROXY: "{{ .Values.proxy.httpsProxy }}"
   NO_PROXY: "{{ template "harbor.noProxy" . }}"
   {{- end }}
+  LOG_LEVEL: "{{ .Values.logLevel }}"
   HARBOR_EXPORTER_PORT: "{{ .Values.metrics.exporter.port }}"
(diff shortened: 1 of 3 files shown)


ywk253100 (approved):

reasonerjt:
If the intention of this PR is solely to make sure the certificate of `harbor-core` pod is trusted by exporter

1. Please refine the commit message to clarify
2. This commit should only make sure the CA is mounted to exporter pod?

templates/internal/auto-tls.yaml
```diff
@@ -95,4 +95,21 @@ data:
   tls.key: {{ $trivyCrt.Key | b64enc | quote }}
 {{- end }}
 
+{{- if and .Values.metrics.enabled}}
+---
+{{- $exporterCN := (include "harbor.exporter" .) }}
+{{- $exporterCrt := genSignedCert $exporterCN nil (list $exporterCN) 365 $ca }}
```
reasonerjt, line 101: I assume we don't need this?

reasonerjt:
Please clarify in the commit message what you are trying to solve in this PR.  `cert issue` is extremely confusing to me.

What you should do is mount the CA in exporter when internalTLS is enabled such that the `exporter` can call the API of `core`.

reasonerjt (approved):

PR #954, opened 2021-05-23
store the file htpasswd in a separate secret

### LINKED ISSUE #866: REGISTRY_HTPASSWD var triggers warning
`REGISTRY_HTPASSWD` seems to be used to add a secret to the registry pod as a file; but it is also noticed by the registry container; which warns `Ignoring unrecognized environment variable REGISTRY_HTPASSWD`; which is a rather concerning message if it makes you think auth isn't being setup.

Can the variable name be changed?

reasonerjt:
fixes #866

### DIFF
--- a/templates/registry/registry-dpl.yaml
+++ b/templates/registry/registry-dpl.yaml
@@ -213,7 +213,7 @@ spec:
       volumes:
       - name: registry-htpasswd
         secret:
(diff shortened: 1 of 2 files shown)


ywk253100 (approved):

ninjadq (approved):
lgtm

PR #974, opened 2021-06-18
fix permission after emergency shutdown kube workers

dopesickjam:
Pod with database cannot start after emergency shutdown kubernetes workers when using longhorn


wirwolf (approved):

ywk253100:
Why the database pod cannot start with longhorn? Is this specific for longhorn?

st0ner:
> Why the database pod cannot start with longhorn? Is this specific for longhorn?

No, 
there is no problem at the first start.
But, when volume remounted - initContainer set chmod for folder - not content, so postgresql cannot access data.
flag '-R' set recursive permissions for folder and content after emergency umounted volume

jindrichskupa:
We had the same issue on EKS with EBS in AWS.

PR #980, opened 2021-06-22
Closes #979 and implements schedulerName for elements that use persistent storage.

### LINKED ISSUE #979: [feature request] support for configuring a different kubernetes scheduler
### Overview
Some storage solutions such as Portworx implement their own scheduler to make sure pods and volumes are colocated. A schedulerName value should be present in the values.yaml to allow this configuration in each of Harbor that uses persistent storage.

Jramirezg:
Closes #979 and implements schedulerName for: 

- trivy
- core
- registry
- redis
- jobservice
- chartmuseum
- database
- exporter


values.yaml
```diff
@@ -423,6 +423,8 @@ core:
     tag: dev
   # set the service account to be used, default if left empty
   serviceAccountName: ""
+  # Use this parameter to set the schedulername for this element. Useful to use Portworx with Harbor
+  schedulerName: ""
```
ywk253100, line 427: How about introducing a global configuration item for the scheduler rather than configuring for all of the pods?

Jramirezg: Agreed. New Pull rewquest created with the new changes to the code. This time the values file only has one new parameter to set the schedulerName globally. Every template reads that parameter.

Jramirezg:
Closed since implemented in another pull request

PR #989, opened 2021-06-28
Add support for ALB ingress controller

kwangil-ha:


ninjadq:
Hi is this PR solve the same issue as #914 ?

templates/ingress/ingress.yaml
```diff
@@ -1,29 +1,31 @@
...
-  {{- $_ := set . "portal_path" "/" -}}
+  {{- $_ := set . "api_path" "/api/*" -}}
+  {{- $_ := set . "service_path" "/service/*" -}}
+  {{- $_ := set . "v2_path" "/v2/*" -}}
+  {{- $_ := set . "chartrepo_path" "/chartrepo/*" -}}
+  {{- $_ := set . "controller_path" "/c/*" -}}
+  {{- $_ := set . "portal_path" "/*" -}}
+  {{- $_ := set . "notary_path" "/*" -}}
+{{- end }}
+{{- else }}
```
democracytoday, line 22: There are now two else-blocks here. Is one of them intended to read:
`{{- else if eq .Values.expose.ingress.controller "alb" }}` ?
This will not be evaluated by Helm correctly as is.

kwangil-ha: Removed two else-blocks in the commit below.
If kubernetes version >= 1.19, path contains postfix "/" and pathType is "Prefix".
otherwise, path contains postfix "*" and pathType is "ImplementationSpecific".

PR #1003, opened 2021-07-07
Allows db service to be exposed via external DNS

mothershipper:
* Adds option for a headless db service
* Adds option for db service annotations

A user can expose the single pod db via external DNS by setting
headless to true and adding the external-DNS annotations to the service.

Signed-off-by: Jack Wink <57678801+mothershipper@users.noreply.github.com>

### DIFF
--- a/templates/database/database-svc.yaml
+++ b/templates/database/database-svc.yaml
@@ -3,11 +3,18 @@ apiVersion: v1
 kind: Service
 metadata:
   name: "{{ template "harbor.database" . }}"
+  annotations:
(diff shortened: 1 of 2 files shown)


ninjadq:
Hi, why we need to expose DB with headless? could tell us more about your use case?

mothershipper:
We're looking to set up db replication to have a hot-standby instance. We're using EKS security groups for pods to control network access between the two regions to avoid paying for a unique LB just to enforce security groups.

Issue #946, opened 2021-05-11
Ingress - Explicit rules order cause incorrect behaviour with AWS ALB using aws-load-balancer-controller


terghalin:
Current harbor-helm chart (2.2.0 and higher) has a root path rule for harbor-portal on top of others, which makes it a priority rule while using AWS ALB.

```
    - path: /*
      backend:
        service:
          name: harbor-{fqdn-name}-harbor-portal
          port:
            number: 80
...
    - path: /c/*
      backend:
        service:
          name: harbor-{fqdn-name}-harbor-core
          port:
            number: 80
```

According to AWS documentation, "_Rules are evaluated in priority order, from the lowest value to the highest value_".

I think it's better to add specific ALB template or just move harbor-portal rule to the bottom, no other controller types should be affected: https://github.com/goharbor/harbor-helm/blob/master/templates/ingress/ingress.yaml

ywk253100:
There is a PR introducing the support for AWS ALB https://github.com/goharbor/harbor-helm/pull/914

jgregmac:
PR [#914](https://github.com/goharbor/harbor-helm/pull/914) addresses path syntax quirks with the AWS Load Balancer Controller in v2.1.3 and earlier.  It does not address the path priority issue noted above.  
Unfortunately, shuffling the harbor-{fqdn-name}-harbor-portal rule to the bottom of the ingress definition does not resolve this problem for me.  My /api/ and /c/ calls still seem to be getting routed to the wrong pod.  This is really frustrating.

democracytoday:
PR #1006 address this explicitly without adding any features specific to ALBs. The change in this PR is not included in #914 and would be required for #914 to work correctly as well.

PR #1006 allows use of the existing Ingress resources with the v2.x series of the AWS Load Balancer Controller without any other changes.

democracytoday:
It's also worth noting that in the meantime you can emulate the functionality of PR #1006 with the following Helm post renderer, here specified as part of a GitOps Toolkit / Flux2 HelmRelease:

```yaml
        patchesJson6902:
          # Reorder Ingress rules to work with AWS Load Balancer
          # Moves all-matching '/*' rule to lowest priority position
          # Can remove when merged: https://github.com/goharbor/harbor-helm/pull/1006
          - target:
              group: networking.k8s.io
              version: v1
              kind: Ingress
              name: harbor-ingress
            patch:
              - op: replace
                path: /spec/rules/0/http/paths/0/path
[...]

PR #1006, opened 2021-07-09
fix(ingress): Reorder portal path for AWS ALBs

democracytoday:
AWS Load Balancer rules are evaluated in priority order, highest to
lowest from top to bottom. This change moves the all-matching / prefix
rule to the bottom of the list to unmask the more specific core paths.
This allows use of an AWS Load Balancer with Harbor.

Previously, all requests, even those matching core paths like /v2, would
be routed to the portal service because the portal path was evaluated as
the highest priority rule.

This address Issue https://github.com/goharbor/harbor-helm/issues/946

### DIFF
--- a/templates/ingress/ingress.yaml
+++ b/templates/ingress/ingress.yaml
@@ -67,10 +67,6 @@ spec:
   - http:
       paths:
 {{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
-      - path: {{ .portal_path }}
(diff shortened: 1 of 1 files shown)


zyyw:
dup of: https://github.com/goharbor/harbor-helm/pull/1132

democracytoday:
You do realise I wrote this 8 months earlier...

democracytoday:
Not sure what's going on with project management here.

PR #1029, opened 2021-07-30
Make IPv6 support opt-in to avoid crashing nginx

funkypenguin:
Hey guys!

A recent commit (https://github.com/goharbor/harbor-helm/commit/a871a2212cef8c0d224f497613b1c1ced6f6ec45) caused nginx to attempt to listen on both IPv4 and IPv6. In the case of a hardened system with IPv6 disabled at the kernel level, this will cause nginx to simply crash, as follows:

```
2021/07/30 02:31:44 [emerg] 1#1: socket() [::]:8080 failed (97: Address family not supported by protocol)
nginx: [emerg] socket() [::]:8080 failed (97: Address family not supported by protocol)
```

Since this may represent a breaking change in environments (_like ours_) where IPv6 is intentionally disabled, this PR allows the user to explicitly enable IPv6 if they need it, but leaves it disabled by default, for backwards compatibility.

Related:

* https://bugs.launchpad.net/ubuntu/+source/nginx/+bug/1743592
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=942817

Cheers!
D


funkypenguin:
Can I get a maintainer review please? This PR fixes a breaking change which could be easily affect many other users like myself..

ninjadq:
hi @funkypenguin, I prefer a more general name like `ipv6.enable`. Or put ipv6 into each component's config like `portal.ipv6` and `nginx.ipv6`. Because `nginxIpv6` isn't general enough to put to the first level of config.
could you update your PR?

ninjadq:
close this PR as per it resolved in https://github.com/goharbor/harbor-helm/pull/1032

Issue #1021, opened 2021-07-23
Add configuration for env CONFIG_OVERWRITE_JSON in harbor-core


stonezdj:
There is a declarative configure setting in harbor-core 
https://goharbor.io/docs/2.3.0/install-config/configure-user-settings-cli/
We should add this configure in Values.yaml

democracytoday:
For the time being you can add a helm post renderer to inject the env var from a secret with the following kustomize patch:
```yaml
          - kind: Deployment
            apiVersion: apps/v1
            metadata:
              name: harbor-core
            spec:
              template:
                spec:
                  containers:
                    - name: core
                      env:
                        - name: CONFIG_OVERWRITE_JSON
                          valueFrom:
                            secretKeyRef:
                              key: CONFIG_OVERWRITE_JSON
                              name: "harbor-config"
```
and Secret:

```yaml
apiVersion: v1
kind: Secret
metadata:
    name: harbor-config
type: Opaque
data:
    # Configure Harbor settings as a single JSON object containing values for the `/api/v2.0/configurations` endpoint.
    # { "auth_mode": "oidc_auth", "oidc_auto_onboard": "true", ... }
    CONFIG_OVERWRITE_JSON: ey...
```

https://helm.sh/docs/topics/advanced/#post-rendering
https://github.com/thomastaylor312/advanced-helm-demos/tree/master/post-render

My example is part of a HelmRelease from the gitops toolkit / flux2 project, but the patch can be applied with just helm and kustomize in the manner described in the advanced-helm-demos repo linked above.

https://fluxcd.io/docs/components/helm/helmreleases/#post-renderers

sorensonpragmatics:
Was going to make exactly this issue and found someone else ran into the same issue.

reasonerjt:
Exposing this in values.yaml doesn't make sense.  We don't expect the user to manually set this value.
This comment
https://github.com/goharbor/harbor-helm/issues/1021#issuecomment-886790784
provides a good example to inject the value via post rendering mechanism, in this case the env var can be injected directly to the pod, so we still dont' need to expose it via `values.yaml`

Therefore, after discussion with @stonezdj, I'm closing this issue as a won't fix

lukasmrtvy:
@reasonerjt will You accept PR implementing setting additional env, envFrom ? Thanks not possible today and this would allow us to set CONFIG_OVERWRITE_JSON  env variable via helm chart directly. Thanks

[...]

PR #1030, opened 2021-08-09
Config overwrite json support

sorensonpragmatics:
This pull request aims to allow helm chart users to be able to specify CONFIG_OVERWRITE_JSON from within `values.yaml`.

See issue https://github.com/goharbor/harbor-helm/issues/1021


reasonerjt:
See my comment in #1021 
Exposing the values.yaml will not be supported because we don't expect the users to set the value manually.

PR #1033, opened 2021-08-16
Added revisionHistoryLimit to all deployments

jaxels10:
I have added revisionHistoryLimit to the following deployments: 

- chartmuseum
- core
- exporter
- jobservice
- nginx
- portal
- registry

I have also added this into the table in the README and added it to values.yaml.

I have also correct a small indentation mistake for the exporter, which was indented with 4 spaces instead of the 2 which is the convention in the rest of the values.yaml file. 

Signed-off-by: Jesper Axelsen <jesperbaxelsen@gmail.com>


jaxels10:
> hi @jaxels10 could you tell why you need to add `revisionHistoryLimit` to values yaml

Hi @ninjadq, 

I mean, there is a general argument to be made about making the deployments more customizable and give the power to the user to decide the variables of the deployment instead either omitting them and letting Kubernetes defaults rule or even defining them and not templating them to the values file. In my opinion, we should aim to give the users of Helm charts as much configurability as possible. I understand that this is not always possible and that Kubernetes manifests have a large number of specs you can define, which could potentially bloat the values.yaml file of a Helm chart, especially if your Helm chart just needs the standard Kubernetes values anyway. 

In a more specific sense, our cluster right now creates a new replicaSet in each deployment every day. That means that 11 days after deploying harbor, we will have 11 x chartmuseum, core, jobservice, notary-server, notary-signer and registry. These revisions are shown in the argoCD UI and I believe it should be customizable how many revisions any Helm chart should save. There could be use cases where people would like to have more revisions saved or fewer.

jacobcrawford:
Oh I really need this feature!

PR #1038, opened 2021-08-25
Allow GCS Workload Identity Authentication For Registry

sharkymcdongles:
The GCS registry driver supports workload idenity if the keyfile is unspecified. I tested this and it works fine. This allows you to specify no keyfile and instead allows application default credentials to authenticate for you with workload identity.

### DIFF
--- a/templates/registry/registry-cm.yaml
+++ b/templates/registry/registry-cm.yaml
@@ -36,7 +36,9 @@ data:
       {{- else if eq $type "gcs" }}
       gcs:
         bucket: {{ $storage.gcs.bucket }}
+        {{- if $storage.gcs.encodedkey }}
         keyfile: /etc/registry/gcs-key.json
+        {{- end }}
         {{- if $storage.gcs.rootdirectory }}
         rootdirectory: {{ $storage.gcs.rootdirectory }}
         {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -281,7 +281,8 @@ persistence:
     gcs:
       bucket: bucketname
       # The base64 encoded json file which contains the key
-      encodedkey: base64-encoded-json-key-file
+      # If encodedkey is unset and GCS enabled it defaults to service account authentication for workload identity
+      #encodedkey: base64-encoded-json-key-file
       #rootdirectory: /gcs/object/name/prefix
       #chunksize: "5242880"
     s3:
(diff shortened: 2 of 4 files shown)


sharkymcdongles:
More info here if curious:

https://github.com/docker/docker.github.io/blob/master/registry/storage-drivers/gcs.md

ninjadq:
could you rebase your pr?

sharkymcdongles:
@ninjadq rebased

templates/registry/registry-cm.yaml
```diff
@@ -36,7 +36,9 @@ data:
       {{- else if eq $type "gcs" }}
       gcs:
         bucket: {{ $storage.gcs.bucket }}
+        {{- if $storage.gcs.encodedkey }}
         keyfile: /etc/registry/gcs-key.json
```
ywk253100, line 40: Seems the `keyfile` here is useless, the volume used in deployment is created from secret. If it is, please remove this line

values.yaml
```diff
@@ -281,7 +281,8 @@ persistence:
     gcs:
       bucket: bucketname
       # The base64 encoded json file which contains the key
-      encodedkey: base64-encoded-json-key-file
+      # If encodedkey is unset and GCS enabled it defaults to service account authentication for workload identity
```
ywk253100, line 284: Please also update the `readme.md` for this property

templates/registry/registry-cm.yaml
sharkymcdongles: `keyfile ` path is only useless if you are using workload identity. This is why if you specify an encoded key, the `keyfile` is still set otherwise it would fail to work when using the encoded key.

values.yaml
sharkymcdongles: I am not sure on the best way to add it to the README since it is a sub value of persistence.imageChartStorage.type that is already covered with 1 line for all the types this is why I added a comment in the values file for it. Thoughts?

`| `persistence.imageChartStorage.type`                                        | The type of storage for images and charts: `filesystem`, `azure`, `gcs`, `s3`, `swift` or `oss`. The type must be `filesystem` if you want to use persistent volumes for registry and chartmuseum. Refer to the [guide](https://github.com/docker/distribution/blob/master/docs/configuration.md#storage) for more information about the detail `

PR #1044, opened 2021-09-03
Add jobservice redis namespace variable

undassa:
Provide ability to set Redis namespace for Job Service


Vad1mo (approved):

ninjadq:
hi @undassa , could you tell what your scenario of using redis namespace?

Vad1mo:
We need the namespace as we are using an external redis service which is used by multiple Harbor instances.

Issue #1045, opened 2021-09-05
Workload's labels doesn't follow the recommendation of labels from Kubernetes.


JhoLee:
tl;dr
Suggesting Harbor Helm Chart to follow the [Kubernetes Labels Recommendation](backward compatibility).
---
I'm using Harbor and many apps with helm. 
And I use their kubernetes labels usefully like;
 1) Collecting metrics with Prometheus.
 2) Managing workloads.

But I knew that Harbor Helm has not follwed Kubernetes' label recommendation. 
([More information](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/))

So I suggest Harbor to follow this recommendation. 
Or I will commit it by myself, and I'll request for review.

Thank you.

ninjadq:
Hi @JhoLee , thank you for mention this, could you help to commit a PR to add this label to harbor-helm, we already have some similar labels but not follow the k8s recommendation, you should keep them and add the new ones. Because some user may depend those labels in production

JhoLee:
@ninjadq  OK. I'll commit it with backward compatibility.

JhoLee:
@ninjadq  I committed those labels with backward compatibility!

martinnirtl:
Hello!! Any update on this?

billabongrob:
Yeah, curious as well.  Seems like none of the deployments or statefulsets inherit any of these labels.  😢

PR #1051, opened 2021-09-21
Add recommended labels for each components

### LINKED ISSUE #1471: Use Standard Helm Labels/Annotations
This Helm chart is not compliant with Helm Label and Annotations.

## Recommendation 
`helm.sh/chart: NAME-VERSION` as a label so that operators can conveniently find all the instances of a particular chart to use.

## Standard Labels

Name | Status | Description
-- | -- | --
app.kubernetes.io/name | REC | This should be the app name, reflecting the entire app. Usually {{ template "name" . }} is used for this. This is used by many Kubernetes manifests, and is not Helm-specific.
helm.sh/chart | REC | This should be the chart name and version: {{ .Chart.Name }}-{{ .Chart.Version \| replace "+" "_" }}.
app.kubernetes.io/managed-by | REC | This should always be set to {{ .Release.Service }}. It is for finding all things managed by Helm.
app.kubernetes.io/instance | REC | This should be the {{ .Release.Name }}. It aids in differentiating between different instances of the same application.
app.kubernetes.io/version | OPT | The version of the app and can be set to {{ .Chart.AppVersion }}.
app.kubernetes.io/component | OPT | This is a common label for marking the different roles that pieces may play in an application. For example, app.kubernetes.io/component: frontend.
app.kubernetes.io/part-of | OPT | When multiple charts or pieces of software are used together to make one application. For example, application software and a database to produce a website. This can be set to the top level application being supported.

https://helm.sh/docs/chart_best_practices/labels/#standard-labels

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

JhoLee:
Add recommended labels for each components of Harbor, and common recommended labels by _helpers.tpl

Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
 
Resolves #1471 #1045

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -4,18 +4,18 @@ metadata:
   name: {{ template "harbor.core" . }}
   labels:
 {{ include "harbor.labels" . | indent 4 }}
-    component: core
+{{ include "harbor.core.labels" . | indent 4 }}
 spec:
   replicas: {{ .Values.core.replicas }}
   selector:
     matchLabels:
 {{ include "harbor.matchLabels" . | indent 6 }}
-      component: core
+{{ include "harbor.core.matchLabels" . | indent 6 }}
   template:
     metadata:
       labels:
 {{ include "harbor.matchLabels" . | indent 8 }}
-        component: core
+{{ include "harbor.core.matchLabels" . | indent 8 }}
       annotations:
         checksum/configmap: {{ include (print $.Template.BasePath "/core/core-cm.yaml") . | sha256sum }}
         checksum/secret: {{ include (print $.Template.BasePath "/core/core-secret.yaml") . | sha256sum }}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -27,10 +27,9 @@ If release name contains chart name it will be used as a full name.
 
 {{/* Helm required labels */}}
 {{- define "harbor.labels" -}}
+{{ include "harbor.matchLabels" . }}
 heritage: {{ .Release.Service }}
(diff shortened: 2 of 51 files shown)


templates/core/core-dpl.yaml
```diff
@@ -4,18 +4,18 @@ metadata:
   name: {{ template "harbor.core" . }}
   labels:
 {{ include "harbor.labels" . | indent 4 }}
-    component: core
+{{ include "harbor.labels" . | indent 4 }}
```
dioguerra, line 7: duplicated?

JhoLee: Oops, typo. It had to be:
```suggestion
{{ include "harbor.core.labels" . | indent 4 }}
```

JhoLee: Fixed it.

JhoLee:
Any updates?

sidewinder12s:
FYI, Changing selector labels will require forced redeployment of all components as that field is immutable, so it'll be a semi-breaking change. Might want to look at not doing that.

zyyw:
could you please squash the commits first and then rebase the code as well. @JhoLee

zyyw:
@sidewinder12s why would it be a concern since this PR is backward compatiable.

sidewinder12s:
@zyyw As far as I'd seen, you cannot change the labels used in selectors after a deployment, it'll error out saying its a immutable field. So if this lands, users cannot deploy it without either force applying it or tearing down their existing deployments. 

https://stackoverflow.com/questions/67679922/how-to-remove-a-label-selector-from-a-kubernetes-deployment

PR #1053, opened 2021-09-27
Use fullname instead of name for ServiceMonitor metadata.name

acobaugh:
When naming a release something like `harbor-foo`, the ServiceMonitor was still being created with `metadata.name` = `harbor`. This should fix that.


ywk253100 (approved):

ywk253100:
@acobaugh Please make sure the DCO checking pass

acobaugh:
DCO check passes now.

PR #1056, opened 2021-09-30
Use "htpasswd" function instead to generate the credential of registry

### LINKED ISSUE #719: htpasswd can be generated with helm
He documentation for `registry.credentials.htpasswd` is 

> The content of htpasswd file based on the value of registry.credentials.username registry.credentials.password. Currently helm does not support bcrypt in the template script, if the credential is updated you need to manually generated by calling htpasswd: htpasswd -nbBC10 $username $password. More details see official_docs.

Helm 3 at least supports bcrypt, see. the https://helm.sh/docs/chart_template_guide/function_list/#htpasswd

This means that `registry.credentials.htpasswd` in https://github.com/goharbor/harbor-helm/blob/42e1887f914e15d54538d03534a038eb43896966/templates/registry/registry-secret.yaml#L9 could be replaced with:

```yaml
{{ htpasswd .Values.registry.credentials.username .Values.registry.credentials.password  | b64enc | quote }} 
```

ywk253100:
Helm v3 introduces the support for template function "htpasswd", use it instead to generate the credential for registry

Fixes #719

Signed-off-by: Wenkai Yin(尹文开) <yinw@vmware.com>


Vad1mo (commented):
Awesome

PR #1072, opened 2021-10-21
feat: add cluster domain

PG2000:
- flatten noProxy chain
- remove unused service names from noProxy
- added cluster domain
- change raw service name to dns service names for appropriate dns resolution within a cluster


PG2000:
@ywk253100 any chance to go forward with this issue?

PR #1073, opened 2021-10-21
Add option to not expose Harbor.

Vad1mo:
This contribution allows users to not expose Harbor so that they use their preferred way of exposing Harbor, that should be outside the scope of this Helm Chart.

The main reasons someone would not expose Harbor is because they are planning to expose Harbor in a way not supported by this Helm Chart:

Examples:

- Istio Gateway
- AWS ALB

Those two examples that have been requested frequently by the community.  
This contribution would allow users to define whatever way they want to expose Harbor.

Closes #1132, #1006, #914, #295, #1672


ninjadq:
hi @Vad1mo , because the error is caused by `commonName`, how about just assign an arbitrary name for `expose.tls.auto.commonName` as a workaround? so we can avoid changes a lots of code here

Vad1mo:
Having this switch explicitly in helm chart has a few advantages over implicity workarounds:

1. Implicit workarounds can stop working and there is no clear documentation or lookup possibility in the code.
2. Explicit documentation and workflow allow users to find exactly the solution to their problem.

Apart from implicit vs. explicit, this feature has a few advantages: 
This functionality allows to use harbor with every ingress controller and harbor helm chart doesn't have to support every use case.

CloudlightCoUk:
I share the use case above. For my project I don't want to use nginx as a proxy. The chart as it is either creates a nginx object, or trys to deploy nginx. I am using Contour with a HTTPProxy object for my front end proxy. The Harbor application architecture defines a proxy layer, so we should have the option to disable this in the chart to 'bring your own proxy'

Vad1mo:
closed in favor https://github.com/goharbor/harbor-helm/pull/1687

PR #1106, opened 2021-12-05
Make postgresql work with Istio

funkypenguin:
Signed-off-by: David Young <davidy@funkypenguin.co.nz>

### DIFF
--- a/.github/workflows/on-push-master-publish-chart.yml
+++ b/.github/workflows/on-push-master-publish-chart.yml
@@ -1,0 +1,79 @@
+name: Publish helm chart
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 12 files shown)


Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 1.4.0-dev
+version: 1.4.0-dev1
```
ninjadq, line 3: we didn't need to change this version

.github/workflows/on-push-master-publish-chart.yml
```diff
@@ -0,0 +1,79 @@
+name: Publish helm chart
```
ninjadq, line 1: this workflow seems not related to you PR

PR #1043, opened 2021-09-03
add className

jetersen:
should allow ingress class resource to work for nginx ingress 👍

PR #1107, opened 2021-12-06
Add ingressClassName to Ingresses

funkypenguin:
Hey guys,

I noticed when deploying the chart into a fresh cluster, that we're not specifying an ingress class using either annotations or ingressClassName. 

This PR adds optional support for specifiying an ingressClassName. I've not bumped Chart.yaml (but am happy to do so if required), because I'm not familiar with your release process :)

Cheers!
D


ywk253100:
Dup of https://github.com/goharbor/harbor-helm/pull/1043.

I'm closing this one @funkypenguin Thanks anyway

PR #1123, opened 2021-12-22
Ability to add htppasswd string instead of credentials.username and credentials.password

identw:
I have a problem using harbor chart with tools like argocd and flux. Due to the htpasswd function being used https://github.com/goharbor/harbor-helm/blob/master/templates/registry/registry-secret.yaml#L44. Each time the template is rendered, there is a new line for REGISTRY_HTPASSWD (since it uses random salt). Therefore, the registry is restarted on every sync in argocd = (.

image

You can avoid this by adding a static line for htpasswd in values.yaml. My edits in this PR make it possible


ninjadq (approved):
lgtm

StevenReitsma:
We're having the same issue. Would be nice if this could get merged!

StevenReitsma:
@ninjadq can this be merged?

ninjadq:
cc: @reasonerjt

danielpacak (approved):

PR #1126, opened 2021-12-27
Existingsecretname

Atoms:
We use GitOps way to manage our deployments with ArgoCD and this is unacceptable to store secrets in git as ArgoCD uses values as it is. Instead, we use an external-secrets operator, which syncs secrets for us, so to be able to manage this proper way I introduced some variables for services where to look for secrets.  This should fix: https://github.com/goharbor/harbor-helm/issues/189 

This can be optimized to use just one secret, for now, it's backward compatible also.


Atoms:
ok, found out some configurations where it's not working, and for now trying to figure out proper fix for that.

Atoms:
As it's harder than at first it looked i will close this, and if i come up with proper way to manage all secrets remotely i will come back

PR #1140, opened 2022-01-25
Add pre-upgrade helm hook job

stonezdj:
The job will migrate database before upgrade
  Add settings of enableMigrateHelmHook to helm chart
  Update readme.md to add the enableMigrateHelmHook option


templates/core/core-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,173 @@
+{{- if .Values.core.enableMigrateHelmHook }}
```
reasonerjt, line 1: Many attributes in the job are copied from the core pod, but are not needed, they should be removed?  IMO the job should be much simpler.

stonezdj: Updated

templates/core/core-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+    {{- with .Values.core.affinity }}
+      affinity:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+    {{- with .Values.core.tolerations }}
+      tolerations:
+{{ toYaml . | indent 8 }}
+    {{- end }}
+    {{- if .Values.core.priorityClassName }}
+      priorityClassName: {{ .Values.core.priorityClassName }}
```
reasonerjt, line 71: is this needed?

README.md
```diff
@@ -223,6 +223,7 @@ The following table lists the configurable parameters of the Harbor chart and th
...
+| `core.enableMigrateHelmHook`                                                | Run the migration job via helm hook                                                                                                                                                                                                                                                                                                                                                                                                                                                            | `false`                                                                             |
```
reasonerjt, line 226: I don't think the option should be put under `core`

templates/core/core-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,73 @@
+{{- if .Values.core.enableMigrateHelmHook }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "harbor.core" . }}-job
```
reasonerjt, line 5: call it migration job?

templates/core/core-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+  labels:
+{{ include "harbor.labels" . | indent 4 }}
+    component: core
```
reasonerjt, line 8: is it correct?

templates/core/core-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-5"
+spec:
+  template:
+    metadata:
+      labels:
+{{ include "harbor.matchLabels" . | indent 8 }}
+        component: core-job
+      annotations:
+        checksum/configmap: {{ include (print $.Template.BasePath "/core/core-cm.yaml") . | sha256sum }}
```
reasonerjt, line 21: this checksum was introduced to control the recreation of the pod, so probably does not make sense for the job?

templates/core/core-pre-upgrade-job.yaml
```diff
@@ -0,0 +1,73 @@
...
+        runAsUser: 10000
+        fsGroup: 10000
+{{- if .Values.core.serviceAccountName }}
+      serviceAccountName: {{ .Values.core.serviceAccountName }}
+{{- end -}}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      automountServiceAccountToken: {{ .Values.core.automountServiceAccountToken | default false }}
```
reasonerjt, line 35: This is for calling the api-server?  probably we don't need it in the job? 
@ywk253100 what do you think?

README.md
stonezdj: updated

templates/core/core-pre-upgrade-job.yaml
stonezdj: Updated

templates/core/core-pre-upgrade-job.yaml
stonezdj: change to migrator

templates/core/core-pre-upgrade-job.yaml
stonezdj: removed

templates/core/core-pre-upgrade-job.yaml
stonezdj: removed

templates/core/core-pre-upgrade-job.yaml
stonezdj: remove

ywk253100 (approved):

PR #1147, opened 2022-02-16
Ability to add environment variables to helm-core container

PhilipNelson5:
I am configuring OIDC via the environment variable [CONFIG_OVERWRITE_JSON](https://goharbor.io/docs/2.4.0/install-config/configure-user-settings-cli/#configure-users-settings-using-an-environment-variable). However, the helm chart does not expose a place to add this environment variable. This feature modifies the helm chart to allow users to add additional environment variables to the core deployment container so they can specify `CONFIG_OVERWRITE_JSON`.

With this addition, a values file like this can be used to configure the user settings via an environment variable
```yaml
core:
  containerEnv:
    - name: CONFIG_OVERWRITE_JSON
      value: >
        {
          "auth_mode": "oidc_auth",
          "oidc_name": "Dex",
          "oidc_endpoint": "https://.../dex",
...
        }
```

The environment variable configuration feature was proposed here goharbor/community#158, implemented here goharbor/harbor#14777, and released as part of harbor [v2.3.0](https://github.com/goharbor/harbor/releases/tag/v2.3.0)


values.yaml
```diff
@@ -450,6 +450,8 @@ core:
   affinity: {}
   ## Additional deployment annotations
   podAnnotations: {}
+  ## Additional environment variables
+  containerEnv: []
```
zyyw, line 454: should this be `containerEnv: {}`?

zyyw, line 454: should this be `containerEnv: []`?

zyyw:
I'm afraid we may not expose this in harbor helm. You may use Post Rendering to manifest of your choice.

values.yaml
PhilipNelson5: I think it should be a list based on the API [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#container-v1-core)

> <table>
> <thead><tr><th>Field</th><th>Description</th></tr></thead>
> <tbody>
> <tr><td><code>env</code><br><i><a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#envvar-v1-core">EnvVar</a> array</i><br><b>patch strategy</b>: <i>merge</i><br><b>patch merge key</b>: <i>name</i></td><td>List of environment variables to set in the container. Cannot be updated.</td></tr>
> </tbody>
> </table>

PhilipNelson5:
I know this is not exposed in harbor helm. The purpose of this pull request is to create a place to add environment variables.

PhilipNelson5:
What is the status of reviewing this feature? Is there something else that needs to be addressed?

I don't mean to be annoying if it's still in progress, I'm not sure what kind of timeline to expect, this is my first time contributing.

janesser:
@zyyw @PhilipNelson5 coming here having the need to configure harbor for oidc via helm

looking into present concern, shifting the env-vars from cm to secret might give it better security (at least obfuscation).

What do you think?

values.yaml
janesser: by the way it is injected, wonder how array or map would fit into the listentry expected having
- name
- value

lemme suggest:
```
{{- if .Values.core.containerEnv }}
- name: CONFIG_OVERWRITE_JSON
- value: {{ .Values.core.containerEnv  }}
{{- end }}
```

thereby .Values.core.containerEnv is expected string, otherwise you would probably need yaml2json.

janesser (changes requested):
injection of helm values in deployment.yaml needs adjustment

@PhilipNelson5 any rough idea when you will continue here, otherwise i can propose to fork-the-fork.

values.yaml
PhilipNelson5: I think this is a good idea. In this case I'd suggest changing the name to something like `.Values.core.configOverwriteJson`. Your change limits this feature to only modifying the CONFIG_OVERWRITE_JSON environment variable instead of allowing arbitrary environment variables to be injected. This might be a good thing.

I don't know of any other environment variables someone would want to set. If there are multiple environment variables that make sense to set, then my way makes more sense. If CONFIG_OVERWRITE_JSON is the only one, then your change makes more sense. I'm happy to use your suggestion. @janesser @zyyw what do you think?

PhilipNelson5:
@janesser is this what you are thinking? https://github.com/PhilipNelson5/harbor-helm/commit/003b0b1af9f0cac4f19eed03293d62eaab8a07b1

janesser:
Absolutely! Last but not least the value could go into a secret, which is ref'd like few lines above.

> @janesser is this what you are thinking? [PhilipNelson5@5131b29](https://github.com/PhilipNelson5/harbor-helm/commit/5131b297fb94aff1391e66e00c411dac8620a152)

PhilipNelson5:
Isn't this already possible with the current state of this branch with something like this?

```yaml
core:
  containerEnv:
    - name: CONFIG_OVERWRITE_JSON
        valueFrom:
          secretKeyRef:
            name: harborSecrets
            key: configureUserSettings
```

janesser:
> Isn't this already possible with the current state of this branch with something like this?
> 
> ```yaml
> core:
>   containerEnv:
>     - name: CONFIG_OVERWRITE_JSON
>         valueFrom:
>           secretKeyRef:
>             name: harborSecrets
>             key: configureUserSettings
> ```

i guess it requires a little extra magic over here: https://github.com/goharbor/harbor-helm/blob/master/templates/core/core-secret.yaml

janesser:
@PhilipNelson5 in case of oidc secrets and most other use-cases we have credentials in hand - so yeah, don't offer bad practice like having credentials in configmaps.

PhilipNelson5:
@janesser, ok I'm onboard. So the `core-dpl.yaml` would look like this?

```yaml
{{- if .Values.core.configureUserSettings }}
          - name: CONFIG_OVERWRITE_JSON
            valueFrom:
              secretKeyRef:
                name: "{{ .Values.core.configureUserSettings }}"
                key: CONFIG_OVERWRITE_JSON
{{- end }}
```

If you want to supply user configuration then you create a secret
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: myHarborSecret
type: Opaque
data:
  CONFIG_OVERWRITE_JSON: eyAiYXV0aF9tb2RlIjogIm9pZGNfYXV0aCIsICJvaWRjX25hbWUiOiAiRGV4IiwgIm9pZGNfZW5kcG9pbnQiOiAiaHR0cHM6Ly8uLi4vZGV4IiwgIm9pZGNfY2xpZW50X2lkIjogImRleC1jb250cm9sbGVyLWhhcmJvci1jbGllbnQiLCAib2lkY19jbGllbnRfc2VjcmV0IjogInBhc3N3b3JkIiwgIm9pZGNfc2NvcGUiOiAib3BlbmlkLG9mZmxpbmVfYWNjZXNzLHByb2ZpbGUsZW1haWwiLCAib2lkY192ZXJpZnlfY2VydCI6ICJ0cnVlIiB9Cg==
```

Then put the name of the secret into the helm values
```yaml
core:
  configureUserSettings: myHarborSecret
```

Does this align with what you are thinking?

janesser:
@PhilipNelson5 We are close ;-)

If you would append a line such as

```yaml
CONFIG_OVERWRITE_JSON: {{ .Values.core.configureUserSettings }}
```
to https://github.com/goharbor/harbor-helm/blob/master/templates/core/core-secret.yaml


```yaml
{{- if .Values.core.configureUserSettings }}
         - name: CONFIG_OVERWRITE_JSON
            valueFrom:
            secretKeyRef:
                name: {{ template "harbor.core" . }}
                key: CONFIG_OVERWRITE_JSON
{{- end }}
```
to the deploy.yaml refering the harbor-std-secret.

PhilipNelson5:
> ```yaml
> CONFIG_OVERWRITE_JSON: {{ .Values.core.configureUserSettings }}
> ```

Do we want ` .Values.core.configureUserSettings` to already be base64 encoded or do we want to add this instead

```yaml
CONFIG_OVERWRITE_JSON: {{ .Values.core.configureUserSettings | b64enc | quote }}
```

janesser:
@PhilipNelson5 You have a point about b64enc. Just like you proposed, looks perfectly fine.

BTW you could refresh the PR - depending on your branching, just push to https://github.com/PhilipNelson5/harbor-helm/tree/core-env-vars again.

PhilipNelson5:
@janesser of course. I was just waiting to hear about the b64enc to update the PR. I think this resolves everything we've talked about.

janesser (approved):
Reviewed latest PR with all improvements. Looks good to me.

@zyyw Please have a look.

janesser:
@zyyw hope your points were covered in a best-practice fashion.

any idea when you will be able to review this?

templates/core/core-dpl.yaml
```diff
@@ -85,6 +85,13 @@ spec:
               secretKeyRef:
                 name: "{{ template "harbor.jobservice" . }}"
                 key: JOBSERVICE_SECRET
+{{- if .Values.core.configureUserSettings }}
+          - name: CONFIG_OVERWRITE_JSON
```
mac-chaffee, line 89: Thanks for doing this! I literally just now ran into the same issue that this PR solves!

Since we set `envFrom: - secretRef:`, I think that means CONFIG_OVERWRITE_JSON is always set as an environment variable for the pod. So wouldn't this block be unneeded?

Would that also mean we need to add `{{- if .Values.core.configureUserSettings }}` around the environment variable in the Secret? That way the environment variable isn't automatically set (by `envFrom`) even when the value is empty.

PhilipNelson5: Glad this will help!

Correct, we only want to define `CONFIG_OVERWRITE_JSON` if `.Values.core.configureUserSettings` is set. We need to add the if statement in the core secret. 

>Since we set `envFrom: - secretRef:`, I think that means `CONFIG_OVERWRITE_JSON` is always set as an environment variable for the container. So wouldn't this block be unneeded?

I don't understand what you mean here. Why is  `CONFIG_OVERWRITE_JSON` always set making this block unneeded?

mac-chaffee: "envFrom" works by taking every key in the Secret and automatically inserting it as an environment variable in the container: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables

So no need to explicitly include the env var here; it is already implicitly included by "envFrom".

janesser: > Thanks for doing this! I literally just now ran into the same issue that this PR solves!
> 
> Since we set `envFrom: - secretRef:`, I think that means CONFIG_OVERWRITE_JSON is always set as an environment variable for the container. So wouldn't this block be unneeded?
> 
> Would that also mean we need to add `{{- if .Values.core.configureUserSettings }}` around the environment variable in the Secret? That way the environment variable isn't automatically set (by `envFrom`) even when the value is empty.

The JSON may contain credentials, see above remarks. Preferably let's keep it "secret".

PhilipNelson5: > "envFrom" works by taking every key in the Secret and automatically inserting it as an environment variable in the container: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables
> 
> So no need to explicitly include the env var here; it is already implicitly included by "envFrom".

@mac-chaffee I understand now, you're talking about line [72-76](https://github.com/goharbor/harbor-helm/blob/562f4ae1e2934fceaf23a5957963e71dcb02cf06/templates/core/core-dpl.yaml#L72-L76). I'll remove the explicit inclusion of the environment variable and add the messing if statement to the core secret.

> The JSON may contain credentials, see above remarks. Preferably let's keep it "secret".

@janesser I think your concerns about the secret are still taken care of. The config will stay in the core secret. We just don't need to be explicitly including it in the core deployment. This is handled by the `envFrom` line linked above

mac-chaffee (approved):

PhilipNelson5:
@reasonerjt or @ywk253100, we've been waiting on a review from @zyyw for many days without hearing from them. Do either of you know what's up? Would you be able to take a look at this PR? We'd super appreciate it!

janesser:
@zyyw don't forget about us here. thanks in advance.

jimsnab (approved):

BulatSaif:
@zyyw, can we merge this PR? I am deploying harbor and would like to have this feature.

BulatSaif:
My current workaround is to use bitnami helm chart: 
https://github.com/bitnami/charts/pull/9534

taemon1337:
I would also like to use this feature

PhilipNelson5:
@zyyw I resolved the conflicts with master. Will you please review this PR?

PhilipNelson5:
@ywk253100  it's been months without hearing from @zyyw. I've recently updated this PR so it is up to date with master. Can we get some feedback or timeline on when to expect a review? Thanks in advance!

PhilipNelson5:
@reasonerjt maybe you can help us move this PR forward. I'm not sure what happened but we are ready for a review and revision or approval. If you could help us we would be very thankful!

0x6b-dev:
@zyyw bump, need this.

Vad1mo (approved):
I am all in with that

igor-nikiforov:
@Vad1mo sorry for ping. Any timeline when this PR will be merged? It is really pain to use this chart without `CONFIG_OVERWRITE_JSON`.

zyyw (approved):

PR #1132, opened 2022-01-12
Move portal at the end of the ingress paths

sathieu:
According to the ingress specification [1], the longest matching path should
be used. But for Istio, "The first rule matching an incoming request is
used" [2],[3].

[1]: https://kubernetes.io/docs/concepts/services-networking/ingress/#multiple-matches
[2]: https://istio.io/latest/docs/reference/config/networking/virtual-service/#VirtualService
[3]: https://github.com/istio/istio/issues/35033

Fixes: #485
Signed-off-by: Mathieu Parent <math.parent@gmail.com>

PR #1148, opened 2022-02-21
Fix Ingress ordering where non-root (non-portal) requests are being routed to portal service

sbhrule15:
This fixes the 405:Method Not Allowed bug, primarily with ALB Ingress controller on Kubernetes >1.19.


zyyw:
dup of https://github.com/goharbor/harbor-helm/pull/1132 ?

sbhrule15:
@zyyw yep! my apologies, didn't see this other PR in my initial review. TBD on merging that other one in?

PR #1159, opened 2022-03-15
Add custom Notary external database credentials

miro-grapeup:
Additional logic in case of different username/password for Notary Server and Signer databases.
If empty value is provided then coreDatabase username/password is used.


zyyw:
Hi @miro-grapeup , I'm afraid we may not merge this PR, because harbor is planning to deprecate and/or remove notary since cosign will be shipped in Harbor 2.5.0

Thank you for your effort.

zyyw:
Closing it because Notary is going to be deprecated.

PR #1165, opened 2022-03-28
Add ability to configure an existing secret for S3 type storage

Sea-you:
In the GitOps world, it's a good practice to separate the configuration of secrets from the actual application Helm chart. 

Using secrets provisioned differently (for example with https://external-secrets.io/) makes it much easier to manage the lifecycle of a secret without the headache of "How would I securely inject my secret with Helm during CD?".

Although this PR only aims to fix that for the `persistence.imageChartStorage.s3` use case, it might encourage others to do the same for other `imageChartStorage` cases.

### DIFF
--- a/templates/chartmuseum/chartmuseum-cm.yaml
+++ b/templates/chartmuseum/chartmuseum-cm.yaml
@@ -69,7 +69,7 @@ data:
   {{- if $storage.s3.regionendpoint }}
   STORAGE_AMAZON_ENDPOINT: {{ $storage.s3.regionendpoint }}
   {{- end }}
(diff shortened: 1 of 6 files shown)


zyyw (approved):
lgtm

Sea-you:
PTAL

zyyw:
Hi @Sea-you, thanks for contributing to harbor-helm, could you please squash your commit into one commit message?

zyyw (approved):

ywk253100 (approved):

frabar:
Is there any chance to see this feature in the stable version soon?

PR #1166, opened 2022-03-28
chore: Configurable caCert and clientCerts validity

imusmanmalik:
CA and Client Certificates have hardcoded 365 days of validity.

This merge request aims to make it configurable.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -138,6 +138,12 @@ internalTLS:
   certSource: "auto"
   # The content of trust ca, only available when `certSource` is "manual"
   trustCa: ""
+  certValidity:
+    # Number of days self-signed certificate is valid
+    caCert: 365
+    # CA certificate validity in days
+    clientCert: 365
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -138,6 +138,12 @@ internalTLS:
   certSource: "auto"
   # The content of trust ca, only available when `certSource` is "manual"
   trustCa: ""
+  certValidity:
+    # Number of days self-signed certificate is valid
+    caCert: 365
+    # CA certificate validity in days
```
zyyw, line 144: the comment of line 144 should be placed before line 143.
the comment of line 146 should be placed before line 145

zyyw:
We would rather keep the self-sign certs with a simple hardcode validity time. If you have strong concerns about certs and security, please use a third-party signed certs.

Closing the PR as for now. Please feel free to open it if you have further questions.

PR #1167, opened 2022-03-30
Corrects comment mistake in Jaeger mode support

### LINKED ISSUE #1164: Change comment typo in Values file around Jaeger Tracing
So I am setting up the Distributed tracing aspect for Harbor and I have followed the helm chart and it was slightly misleading.

The below is the yaml that can be found here:
https://github.com/goharbor/harbor-helm/blob/6a52e9a35ddcffb268aea688766a1a2cff2286da/values.yaml#L910-L919

However i think the `agent mode` and `collector mode` comments are the wrong way around. I have tested this because when I point my `endpoint` at my jaeger collector: `endpoint: http://jaeger-collector.jaeger:14268/api/traces`, I can the traces showing up. However when putting the agent details in the `endpoint` value, I get nothing. I also think the fact that in the `collector mode` it has `agent_host` and `agent_port` which is also a bit of a give away to why I believe it's a typo.

Should be a quick fix, just wanted to double check with someone first.

ChrisJBurns:
Corrects the comment mistake in the Jaeger support mode that has the `agent` and `collector` mode description the wrong way around.

Fixes #1164

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -909,8 +909,8 @@ trace:
   #   application: harbor
   jaeger:
     # jaeger supports two modes:
-    #   agent mode(uncomment endpoint and uncomment username, password if needed)
-    #   collector mode(uncomment agent_host and agent_port)
+    #   collector mode(uncomment endpoint and uncomment username, password if needed)
+    #   agent mode(uncomment agent_host and agent_port)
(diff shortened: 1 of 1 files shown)


zyyw (approved):
lgtm

ywk253100 (approved):

Issue #1084, opened 2021-10-30
harbor-db not starting (initdb: error: could not create directory / Permission denied)


bbqrob:
Using helm chart v1.7.3 (using application v2.3.3) as seen [here](https://artifacthub.io/packages/helm/harbor/harbor), installed the chart into namespace `harbor`.

```
ls: cannot access '/var/lib/postgresql/data/pgdata': No such file or directory
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locales
  COLLATE:  en_US.UTF-8
  CTYPE:    en_US.UTF-8
  MESSAGES: C
  MONETARY: C
  NUMERIC:  C
  TIME:     C
The default text search configuration will be set to "english".

Data page checksums are disabled.

creating directory /var/lib/postgresql/data/pgdata/pg13 ... 2021-10-30T19:16:21.245845654+02:00 initdb: error: could not create directory "/var/lib/postgresql/data/pgdata": Permission denied
```

My `values.yaml` got the following snippet for database claim:

```
    database:
      existingClaim: "database-data-harbor-database-0"
      storageClass: ""
      subPath: ""
      accessMode: ReadWriteOnce
      size: 2Gi
```

PersistentVolume is configured as:

```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: database-data-harbor-database-0
  namespace: harbor
  labels:
    app: harbor-pg
    type: local
spec:
    capacity:
        storage: 2Gi
    accessModes:
        - ReadWriteOnce
    hostPath:
        path: "/var/kubelet/harbor/pg"
```
with claim:

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: database-data-harbor-database-0
  namespace: harbor
  labels:
    app: harbor-pg
spec:
  storageClassName: ""
  capacity:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
  selector:
    matchLabels:
      app: harbor-pg
```

bbqrob:
Adding 

```
subPath: "data" 
```

simply created the directory in the filesystem, but didn't change anything to the permission issue.

zyyw:
currently, we use fsGroup to change the Permission, does your storage provider/driver support fsGroup?
alternately, you can use initContainer to update the file permission with root.

timbrown5:
I managed to work around this with an init container, I used the following YAML

[...]

PR #1170, opened 2022-04-01
Add new init container to database_ss to chown data volume

timbrown5:
I had issues when using an NFS mount point as my volume. I worked around this by `chown`-ing the data mount point as the Postgres user (`999:999`): 

https://github.com/goharbor/harbor-helm/issues/1084#issuecomment-1085600354

### DIFF
--- a/templates/database/database-ss.yaml
+++ b/templates/database/database-ss.yaml
@@ -55,6 +55,20 @@ spec:
           - name: database-data
             mountPath: /var/lib/postgresql/data
             subPath: {{ $database.subPath }}
+      # if the volumn group is an NFS mount-point we need to ensure the postgres user will own the data directory. command
+      - name: "data-ownership-ensurer"
+        image: {{ .Values.database.internal.image.repository }}:{{ .Values.database.internal.image.tag }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        command: ["/bin/sh"]
+        args: ["-c", "chown -R 999:999 /var/lib/postgresql/data/pgdata || true"]
+{{- if .Values.database.internal.initContainer.permissions.resources }}
+        resources:
+{{ toYaml .Values.database.internal.initContainer.permissions.resources | indent 10 }}
+{{- end }}
+        volumeMounts:
+          - name: database-data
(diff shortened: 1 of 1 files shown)


templates/database/database-ss.yaml
```diff
@@ -50,6 +50,20 @@ spec:
...
+          - name: database-data
+            mountPath: /var/lib/postgresql/data
+            subPath: {{ $database.subPath }}
+      # if the volumn group is an NFS mount-point we need to ensure the postgres user will own the data directory. command
+      - name: "data-ownership-ensurer"
+        image: {{ .Values.database.internal.image.repository }}:{{ .Values.database.internal.image.tag }}
+        imagePullPolicy: {{ .Values.imagePullPolicy }}
+        command: ["/bin/sh"]
+        args: ["-c", "chown -R 999:999 /var/lib/postgresql/data || true"]
+{{- if .Values.database.internal.initContainer.permissions.resources }}
```
zyyw, line 64: this `.Values.database.internal.initContainer.permissions.resources` should probably be `.Values.database.internal.initContainer.ownership.resources`, and add it to values.yaml file

templates/database/database-ss.yaml
```diff
@@ -50,6 +50,20 @@ spec:
...
+        resources:
+{{ toYaml .Values.database.internal.initContainer.permissions.resources | indent 10 }}
```
zyyw, line 66: same thing here: `.Values.database.internal.initContainer.permissions.resources`

zyyw:
I'm wondering why only db-ss encountered with the **permission denied** issue. How about
* harbor-redis
* harbor-trivy
* harbor-jobservice
* harbor-registry
There is no **permission denied** issue for the ones above?

zyyw:
We could prefer to modify the permission through fsGroup. 
* if use "chown", it requires root permission to run the init container, which might have security concern.

Closing it for now.

marcusnh:
i'm experiencing this issue where I have permission denied for:

-     harbor-redis
-     harbor-trivy
-     harbor-jobservice
-     harbor-registry

is there a way to workaround this?

PR #1173, opened 2022-04-08
fixed a bug when setting CA bundle

Maven35:
encountered a bug in my environment that when a custom CA bundle was being used Trivy did not pick it up. let me know if this change works or if you guys want to do it a different way.

### DIFF
--- a/templates/trivy/trivy-sts.yaml
+++ b/templates/trivy/trivy-sts.yaml
@@ -109,6 +109,10 @@ spec:
                 secretKeyRef:
                   name: {{ template "harbor.trivy" . }}
                   key: redisURL
+            {{- if .Values.caBundleSecretName }}
+            - name: "SSL_CERT_DIR"
+              value: /harbor_cust_cert
+            {{- end }}
           ports:
             - name: api-server
               containerPort: {{ template "harbor.trivy.containerPort" . }}


templates/trivy/trivy-sts.yaml
```diff
@@ -109,6 +109,10 @@ spec:
                 secretKeyRef:
                   name: {{ template "harbor.trivy" . }}
                   key: redisURL
+            {{- if .Values.caBundleSecretName }}
```
zyyw, line 112: The caBundleSecretName has already been implemented in trivy-sts.yaml:
https://github.com/goharbor/harbor-helm/blob/master/templates/trivy/trivy-sts.yaml#L124-L126
https://github.com/goharbor/harbor-helm/blob/master/templates/trivy/trivy-sts.yaml#L154-L156

templates/trivy/trivy-sts.yaml
```diff
@@ -109,6 +109,10 @@ spec:
...
+            - name: "SSL_CERT_DIR"
```
zyyw, line 113: does it have to be this `SSL_CERT_DIR` in trivy?

Maven35: SSL_CERT_DIR is an env variable for the OpenSSL libraries this overrides the default with the declared vaule in my environment trivy was not picking up and using the CABundle that was being pushed into the container, so this was my work around to fix it not sure if its the best or not but it does work if the cabundlesecretname variable is set.

zyyw (approved):
lgtm

Maven35:
anyone wanna merge this in?

PR #1179, opened 2022-04-20
Add ability to use external secrets

golgoth31:
Hi,
Thank you for this amazing job.
please find here an update of the charts files to allow usage of existing secrets instead of storing secret values inside values file.

regards

David Sabatie

PR #1209, opened 2022-05-28
Add flags to output no secrets, or just secrets

meln5674:
https://github.com/goharbor/harbor-helm/issues/189 has languished for an unacceptably long time, and it appears that pull requests to fix it have stalled out multiple times. Reading both the issue and pull request review comments, it appears this is due in part to the complexity of the chart, and perhaps an impedance mismatch between the desires of the chart users and maintainers.

As such, I am proposing a radically simpler solution: Instead of specifying existing secrets, simply don't create them, and leave the user responsible for creating secrets with the expected names and keys.

In aid of this approach, I have added two fields `gitops.noSecrets`, which removes all Secrets from the helm output, and `gitops.onlySecrets` which removes everything but secrets from the helm output. The intention is that users wishing to use the chart in a GitOps context add `--set gitops.noSecrets=true` (or the equivalent in their GitOps tool of choice) to create the non-sensitive resources, and then, out of band, use `--set gitops.onlySecrets=true` to template out the secrets that the chart expects with the same namespace, release name, and other values. The user can then create these manually with kubectl, or use an encryption solution like https://github.com/bitnami-labs/sealed-secrets . So long as the expected secrets are created, the other chart resources should be none the wiser. This solution is scalable and maintainable, as any new resources added to the chart simply need to be enclosed in a `{{ if }} {{ end }}` block. If these flags are not set, the chart operates as if they did not exist, preserving behavior for any existing chart users.

I should also mention that this is in no way incompatible with any of the in-flight pull requests to allow specifying specific secrets. This is simply a "scorched-earth" approach to guarantee there are no sensitive values required in the input.


zyyw:
we are gradually reviewing PRs to accept specifying existing secrets. I'm afraid this PR will not be merged. 
Thanks for your contribution all the same!!

meln5674:
Are you referring to https://github.com/goharbor/harbor-helm/pull/1179 ? Not only has it sat unchanged for nearly a month (just like the last attempts at a PR), your comment here https://github.com/goharbor/harbor-helm/pull/1179#discussion_r868787339 seems to indicate that not *all* secrets will be allowed to be provided external to the chart. _Any_ secret, whether provided in the values.yaml or randomly generated with sprig, disqualifies a chart for use within gitops, which is a major theme in comments in the linked issue.

I'll check back in 6 months when https://github.com/goharbor/harbor-helm/issues/189 will inevitably still be open.

PR #1242, opened 2022-07-28
htpasswdString doc fix - resolves #1210

### LINKED ISSUE #1210: htpasswdString is broken
`registry.credentials.htpasswdString`                
Is supposed to be an alternative to username password:
           
> Login and password in htpasswd string format. Excludes `registry.credentials.username`  and `registry.credentials.password`. May come in handy when integrating with tools like argocd or flux. This allows the same line to be generated each time the template is rendered, instead of the `htpasswd` function from helm, which generates different lines each time because of the salt.  

However the implementation of `htpasswdString` in the Helm Chart is broken, so if you set `htpasswdString` you can login.

This search result summarizes why it is not working https://github.com/goharbor/harbor-helm/search?q=registry.credentials.password

### It is not working because
1. [templates/jobservice/jobservice-secrets.yaml](https://github.com/goharbor/harbor-helm/blob/46771e20d749395cdeefb4076ec9aad24527c6ac/templates/jobservice/jobservice-secrets.yaml) 
`REGISTRY_CREDENTIAL_PASSWORD: {{ .Values.registry.credentials.password | b64enc | quote }}`
2. [templates/core/core-secret.yaml](https://github.com/goharbor/harbor-helm/blob/46771e20d749395cdeefb4076ec9aad24527c6ac/templates/core/core-secret.yaml) ->
`REGISTRY_CREDENTIAL_PASSWORD: {{ .Values.registry.credentials.password | b64enc | quote }}`

Both Jobservice and Core are still trying to use `registry.credentials.password` which is different. 

The solution would be to allow `registry.credentials.password` together with `registry.credentials.htpasswdString`

RichardSufliarsky:
Signed-off-by: Richard Sufliarsky <richard.sufliarsky@gmail.com>

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -576,8 +576,14 @@ registry:
   credentials:
     username: "harbor_registry_user"
     password: "harbor_registry_password"
-    # Login and password in htpasswd string format. Excludes `registry.credentials.username`  and `registry.credentials.password`. May come in handy when integrating with tools like argocd or flux. This allows the same line to be generated each time the template is rendered, instead of the `htpasswd` function from helm, which generates different lines each time because of the salt.
-    # htpasswdString: $apr1$XLefHzeG$Xl4.s00sMSCCcMyJljSZb0 # example string
+    # Login and password in htpasswd string format. May come in handy when integrating with tools like argocd or flux.
+    # This allows the same line to be generated each time the template is rendered, instead of the `htpasswd` function from helm,
+    # which generates different lines each time because of the salt.
+    # Do not comment out `registry.credentials.username` and `registry.credentials.password` as these are used
+    # for core and jobservice Config Maps and Secrets. 
+    # To generate bcrypt hash: "htpasswd -nbBC10 $username $password".
+    # More info: https://httpd.apache.org/docs/2.4/misc/password_encryptions.html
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -576,8 +576,14 @@ registry:
   credentials:
     username: "harbor_registry_user"
     password: "harbor_registry_password"
-    # Login and password in htpasswd string format. Excludes `registry.credentials.username`  and `registry.credentials.password`. May come in handy when integrating with tools like argocd or flux. This allows the same line to be generated each time the template is rendered, instead of the `htpasswd` function from helm, which generates different lines each time because of the salt.
-    # htpasswdString: $apr1$XLefHzeG$Xl4.s00sMSCCcMyJljSZb0 # example string
+    # Login and password in htpasswd string format. May come in handy when integrating with tools like argocd or flux.
+    # This allows the same line to be generated each time the template is rendered, instead of the `htpasswd` function from helm,
+    # which generates different lines each time because of the salt.
+    # Do not comment out `registry.credentials.username` and `registry.credentials.password` as these are used
```
zyyw, line 582: please add an additional comment:
```
If using htpasswdString, it should be consistent with using credentials.username and credentials.password.
```

RichardSufliarsky: @zyyw, done.

Vad1mo (approved):

PR #1244, opened 2022-07-28
feat: Allow using preexisting secrets for sensitive data

sylvainOL:
There are a lot of sensitive datas (passwords, secrets, keys, ...) that needs to be given.
This patch allows to set them in a secret before installing and using the secret.

It's a new version of #1179 (I've redone it, I didn't reused it).

Current limitations:

* redis password with external may not work (as many components uses the "direct" url and it's not easy to fix that)
* notary has not been touched


sylvainOL:
After review, I think this patch tries to solve too many things...
I'll create simpler one for each secret if you don't mind

PR #1274, opened 2022-08-30
Allow to give admin password via secret

sylvainOL:
Instead of setting the admin password via a parameter in the helm chart,
provide the ability to give it via an existing secret.


sylvainOL:
Hello, I'm sorry to bother you but is there any reason not to at least launch tests? (same question for #1290 and #1291)

sylvainOL:
@zyyw , @MinerYang can you take a look?

README.md
```diff
@@ -196,6 +196,8 @@ The following table lists the configurable parameters of the Harbor chart and th
...
 | `harborAdminPassword`                                                | The initial password of Harbor admin. Change it from portal after launching Harbor                                                                                                                                                                                                                                                                                                                                                                                | `Harbor12345`                         |
+| `existingSecretAdminPassword`                                                | The name of secret whera admin password can be found. Harbor                                                                                                                                                                                                                                                                                                                                                                                |                          |
```
pitwegner, line 199: typo in "where". Also, why the trailing "Harbor"? Same in line below.

templates/core/core-dpl.yaml
```diff
@@ -86,7 +86,14 @@ spec:
               secretKeyRef:
                 name: "{{ template "harbor.jobservice" . }}"
                 key: JOBSERVICE_SECRET
-         {{- if .Values.internalTLS.enabled }}
+          {{- if .Values.existingSecretAdminPassword }}
```
pitwegner, line 89: why not simply add another if-guarded line to the secret? This kind of breaks the convention here.

templates/exporter/exporter-dpl.yaml
```diff
@@ -64,6 +64,13 @@ spec:
               name: {{ .Values.database.external.existingSecret }}
               key: password
         {{- end }}
+        {{- if .Values.existingSecretAdminPassword }}
```
pitwegner, line 67: Same as above. Keep it simple by adding a line to the existing secret.

pitwegner (changes requested):

README.md
sylvainOL: good catch thanks, I've just resolved it!

templates/core/core-dpl.yaml
sylvainOL: Maybe I don't catch your comment.

But if the comment is to set it in "core" secret that is generated, I guess we can't
The goal here is to retrieve the admin password value from a "previously generated" one.
Then, we cannot add it to the core secret.
And then we need to set it this way AFAIK.

please don't hesitate to correct me as I may didn't understood your comment

templates/exporter/exporter-dpl.yaml
sylvainOL: same as previous comment

sylvainOL (commented):
Thanks for the review!

I solved readme issue and tried to answer on the secret part.

templates/core/core-dpl.yaml
pitwegner: Absolutely right. My fault.

rdelpret:
@pitwegner any movement on this? We would really like this feature as well.

pitwegner:
@rdelpret I'm not related to the goharbor project at all, I just added my 2 cents to the PR because I'm interested in the feature as well

rdelpret:
@MinerYang @zyyw Any movement on this? also looking for any help with maintaining this chart?

zyyw:
closing it and then reopen it to run the ci

zyyw (approved):

ywk253100 (approved):

PR #1288, opened 2022-09-05
Harbor Registry -- Fix existing secret reference in deployment

azalenski-branch:
Current code for specifying an existing secret for htpasswd in registry configuration doesn't work as expected due to missing `{{}}` to enable helm templating.

This change fixes this issue by adding `{{}}` to the respective parameter.


MinerYang (approved):
lgtm

PR #1291, opened 2022-09-06
fix(otel): use right default value for timeout

sylvainOL:
According to
[Harbor config](https://github.com/goharbor/harbor/blob/main/src/lib/config/metadata/metadatalist.go#L182),
OTEL timeout has to be set as int.


sylvainOL:
@zyyw , @MinerYang can you take a look?

MinerYang:
Hi @sylvainOL ,
Sorry for the late response. As you can see that the `values.yaml` file configuration is a template that keep consistent with upstream goharbor repository. It use `5s` to make it more clear for users that is takes time using `seconds`. But  is would converted into int type eventually in the code base.
https://github.com/goharbor/harbor/blob/159169227b50b09fcf9553d517f4c6b04acb00b8/src/lib/config/trace.go#L33
https://github.com/goharbor/harbor/blob/159169227b50b09fcf9553d517f4c6b04acb00b8/src/lib/config/metadata/value.go#L67.
So i don't think this change is necessary.
Thanks for your contribution!

Best,
Miner

sylvainOL:
Hello, I created this patch because I saw this message in the log at start when using `10s` and not `10`:

```
2022-10-13T08:31:52Z [INFO] [/pkg/reg/adapter/tencentcr/adapter.go:41]: the factory for adapter tencent-tcr registered
2022-10-13T08:31:52Z [ERROR] [/pkg/config/manager.go:66]: LoadSystemConfigFromEnv failed, config item, key: trace_otel_timeout,  err: invalid int string: 10s
2022-10-13T08:31:52Z [ERROR] [/pkg/config/manager.go:66]: LoadSystemConfigFromEnv failed, config item, key: trace_otel_timeout,  err: invalid int string: 10s
2022-10-13T08:31:52Z [DEBUG] [/core/auth/authenticator.go:133]: Registered authentication helper for auth mode: http_auth
2022-10-13T08:31:52Z [DEBUG] [/core/auth/authenticator.go:133]: Registered authentication helper for auth mode: db_auth
2022-10-13T08:31:52Z [DEBUG] [/core/auth/authenticator.go:133]: Registered authentication helper for auth mode: ldap_auth
2022-10-13T08:31:52Z [DEBUG] [/core/auth/authenticator.go:133]: Registered authentication helper for auth mode: oidc_auth
2022-10-13T08:31:52Z [DEBUG] [/core/auth/authenticator.go:133]: Registered authentication helper for auth mode: uaa_auth
2022-10-13T08:31:52Z [ERROR] [/pkg/config/manager.go:66]: LoadSystemConfigFromEnv failed, config item, key: trace_otel_timeout,  err: invalid int string: 10s
```

when setting `10`, the error disappears.

zyyw:
Closing the PR and then open it again to run the CI.

zyyw (approved):
sorry for the late review and thanks for contributing to harbor-helm!
lgtm

ywk253100 (approved):

Issue #1296, opened 2022-09-14
External Redis Auth failed


MinerYang:
when i install harbor with harbor-helm v1.10 with external Redis, there would be a problem when auth and connect to the external redis server with bellowing error.
```
 2022-09-08T16:18:26Z [ERROR] [/lib/cache/cache.go:114]: failed to ping redis://redis:xxxxx@10.191.160.140:6379/0?idle_timeout_seconds=30, retry after 10s : ERR wrong number of arguments for 'auth' command
2022-09-08T16:18:36Z [ERROR] [/lib/cache/cache.go:114]: failed to ping redis://redis:xxxxx@10.191.160.140:6379/0?idle_timeout_seconds=30, retry after 10s : ERR wrong number of arguments for 'auth' command
```

this issue might cause by wrong Redis url rendered after upstream oss change redis library from `redigo` to `go-redis` since v2.6

https://github.com/go-redis/redis/issues/1343

niklasweimann:
Could you create a v1.10.1 Release including this fix, please? This would be very helpful. Thanks 😍😎

niklasweimann:
@ywk253100 @zyyw @MinerYang  Hey 👋  Any progress on this?

PR #1297, opened 2022-09-14
fix external redis url

MinerYang:
Signed-off-by: yminer <yminer@vmware.com>
https://github.com/goharbor/harbor-helm/issues/1296

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -168,16 +168,16 @@ postgres://{{ template "harbor.database.username" . }}:{{ template "harbor.datab
   {{- end }}
 {{- end -}}
 
-/*scheme://[redis:password@]host:port[/master_set]*/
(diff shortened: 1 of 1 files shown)


templates/_helpers.tpl
```diff
@@ -169,39 +169,40 @@ postgres://{{ template "harbor.database.username" . }}:{{ template "harbor.datab
 {{- end -}}
 
 /*scheme://[redis:password@]host:port[/master_set]*/
```
zyyw, line 171: do we need to remove this line 171 as well?

zyyw:
please also update the commit message to something like "change the username of external redis to empty string instead of redis" to make it more descriptive, thanks.

ywk253100 (approved):

zyyw (approved):

PR #1307, opened 2022-09-21
Allow to configure timeoutSeconds in livenessProbe and readinessProbe

peter-englmaier:
Introduce new values to configure timeoutSeconds in livenessProbe/readinessProbe:

.database.internal.livenessProbe.timeoutSeconds
.database.internal.readinessProbe.timeoutSeconds

Default values (1 second) unchanged if parameters are not set.

In GKE the internal database reports "Readiness probe failed". Setting .database.internal.readinessProbe.timeoutSeconds: 3
fixes the problem.

Please let me know if I need to add documentation for the new values elsewhere.


peter-englmaier:
I have tested the helm chart in gke. After setting timeout to 3 seconds, our instance does no longer get 'readiness probe failed' errors:
image

values.yaml
```diff
@@ -777,6 +777,12 @@ database:
     #  requests:
     #    memory: 256Mi
     #    cpu: 100m
+    # The timeout used in livenessProbe; 1 to 5 seconds
+    livenessProbe:
```
ywk253100, line 781: Could you add the same configuration for all other components so that they are consistent?

peter-englmaier: What other components? I want to touch the internal database only.

peter-englmaier:
> @peter-englmaier could you please update the README.md file to add items of
> 
> ```
> .database.internal.livenessProbe.timeoutSeconds
> .database.internal.readinessProbe.timeoutSeconds
> ```
>
> and then rebase the latest master into your working branch?

OK, done. 
There was a [comment](https://github.com/goharbor/harbor-helm/pull/1307#discussion_r1000302140) which is open.

zyyw (approved):

values.yaml
peter-englmaier: @ywk253100 : what other components to you have in mind? Yes, the probes have more parameters and as a regular harbor user, I also like to have things be consistent.

peter-englmaier: @ywk253100 : Can you please comment what you need for your approval? I'm a bit lost. I can add more configurable parameters, but which ones? My PR is for liveness/readinessProbes for the build-in database only. I can also use other parameter names if you find them not suitable. Best, Peter

ywk253100 (approved):

peter-englmaier:
Thanks @zyyw !

PR #1316, opened 2022-09-30
Add certificate support for PSQL SSL connections

damyan:
Provide client/server certificate data as a kubernetes secret

Note: notary-{server,signer} does not work due to fsGroup + permission restriction, s https://github.com/kubernetes/kubernetes/issues/57923

Signed-off-by: Damyan Yordanov <damyan.yordanov@sap.com>


damyan:
Err

PR #1326, opened 2022-10-17
Allow usage of existingSecret for database

### LINKED ISSUE #1013: Ability to include secret instead of exposing raw password
Exposing database and redis secret via helm values files are very dangerous. It would be better to mount k8s secret.

Proposed change (example)

```
database:
  external:
    existingSecret: "SECRET-NAME"
    existingSecretPasswordKey: "SECRET-KEY"
```

PKizzle:
Instead of using the redis.external.password use `REDIS_PASSWORD` inside `.Values.redis.external.existingSecret` if provided.
If provided uses `username` and `password` in `.Values.database.external.existingSecret` instead of raw password.

Fixes #1013


sylvainOL:
I didn't knew the "lookup" feature of helm, nice use of it :)

PKizzle:
@sylvainOL I also just recently found it online and now am using it quite extensively 😅
I hope you don't mind another question regarding this PR: Right now there are two different schemes for the password inside the secrets. For the database I am using `username` and `password` while Redis requires `REDIS_PASSWORD` (env variable style) since it is used by harbor-helm elsewhere. However the Redis operator I am using requires `password` to be set inside the secret (https://github.com/spotahome/redis-operator). What do you think of another PR that changes all `REDIS_PASSWORD` secret references to `password`?

sylvainOL:
Hi @PKizzle, I'm a "nobody" here, just a guy who has proposed patches to enhance "security" for deployment and look at other patches related to that :)

As far as I can see how use of existing secret is done today, most of the time the name of the secret will be the name of the needed environment variable.

So I guess you should stay with `REDIS_PASSWORD` IMHO

ricardojdsilva87:
Hello,
This would be nice to have, since the only way of making the secret work is to put it in cler text on the values file since it's transformed to the `harbor.database.rawPassword` leading to the composed string for Redis:
https://github.com/goharbor/harbor-helm/blob/master/templates/_helpers.tpl#L139

Since we are using already the secret password for other env variables for other deployments it would be nice to use it also to have the username and password for the database.
Thanks for the great work!

zyyw:
closing this PR and reopen it to trigger the CI

templates/_helpers.tpl
```diff
@@ -164,15 +168,17 @@ postgres://{{ template "harbor.database.username" . }}:{{ template "harbor.datab
 
 {{- define "harbor.redis.password" -}}
   {{- with .Values.redis }}
-    {{- ternary "" .external.password (eq .type "internal") }}
+    {{- if ne .type "internal" -}}
+      {{- ternary ((lookup "v1" "Secret" $.Release.Namespace .external.existingSecret).data.REDIS_PASSWORD | b64dec) .external.password (not (not .external.existingSecret)) }}
```
zyyw, line 172: Is is possible to use the "style" as it is in the original PR:
* https://github.com/goharbor/harbor-helm/pull/1179

instead of using a new function `lookup` in order to keep consistency of the code and for the sake of code maintenance. Thanks

PKizzle: While it is possible to adjust the "style" for `POSTGRESQL_PASSWORD` to the one mentioned in the PR #1179, this does not work for `POSTGRESQL_USERNAME` as a postgres URL has to be built from it and this requires access to the secret's content. For keeping the "style" and not use the lookup function code changes would be required to allow harbor to internally convert the database username into a postgres URL. The other option would be to create a secret containing the whole postgres URL, but that defeats my use-case as I am trying to use an already existing secret containing the database credentials.

templates/_helpers.tpl
```diff
@@ -162,17 +166,27 @@ postgres://{{ template "harbor.database.username" . }}:{{ template "harbor.datab
...
   {{- end }}
 {{- end -}}
 
-/*scheme://[:password@]host:port[/master_set]*/
+/*scheme://[username][:password@]host:port[/master_set]*/
 {{- define "harbor.redis.url" -}}
   {{- with .Values.redis }}
     {{- $path := ternary "" (printf "/%s" (include "harbor.redis.masterSet" $)) (not (include "harbor.redis.masterSet" $)) }}
-    {{- $cred := ternary (printf "%s:%s@" (.external.username | urlquery) (.external.password | urlquery)) "" (and (eq .type "external" ) (not (not .external.password))) }}
+    {{- $cred := ternary (printf "%s:%s@" ((include "harbor.redis.username" $) | urlquery)) ((include "harbor.redis.password" $) | urlquery)) "" (and (eq .type "external" ) (not (not (include "harbor.redis.password" $)))) }}
```
benedikt-bartscher, line 189: helm lint is not happy with this change. You added one `)`to much after the first `urlquery`.

Should look like this:
```
    {{- $cred := ternary (printf "%s:%s@" ((include "harbor.redis.username" $) | urlquery) ((include "harbor.redis.password" $) | urlquery)) "" (and (eq .type "external" ) (not (not (include "harbor.redis.password" $)))) }}
```

Kajot-dev:
This is more of a workaround. I would expect that if i change redis password inside the secret and restart harbor pods, the'll use the new password but that's not the case when using lookup. Additionally this will not work when using `helm template`.

However you can **dynamically construct env variables using other env variables** like this:

```yaml
env:
  - name: REDIS_PASSWORD
    valueFrom:
      secretKeyRef:
        name: {{ .Values.redis.external.existingSecret }}
        key: REDIS_PASSWORD
  - name: _REDIS_URL_CORE
    value: "redis+sentinel://:$(REDIS_PASSWORD)@harbor-redis-headless:26379/harbor-redis-master/0?idle_timeout_seconds=30"
``` 

Reference: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

Kajot-dev:
Yeah, you're right, templating from the ConfigMap does not work unfortunately. Probably moving this to the container's entrypoint would have been a better idea, like this:
- If a complete redis url is present in the environment, use that
- If redis url does not have a password but `REDIS_PASSWORD` is present, inject it into the redis url

And I don't have the issue with the fact that if you want to update the configuration, I must redeploy the chart - I have the issue with it stating that I can use redis password from the existing secret which is simply not true.

PKizzle:
What do you exactly mean by "I can't use redis password from the existing secret"?
I am using this PR for redis and database credentials stored in secrets and it is working fine for me.

Kajot-dev:
Sorry I meant this in a context of the official, unmodified chart, this PR is prefectly fine!

PKizzle:
@sylvainOL @zyyw What do you think would be the best option to move forward?

Should the helm chart be rewritten for redis and database URL construction using dynamic environment variables (for automatic secret updates) or should this happen inside harbor's source code to reduce code/config complexity?

Another option would be to use an init container that uses something like [envsubst](https://linux.die.net/man/1/envsubst) to keep the URL construction outside of any harbor related source code.

sylvainOL:
HI @PKizzle, my preferred way would be to have the "magic" in harbor source code but if it's not possible, the "kubernetes way" (injecting variables in variables) seems the most appropriate.

This would allow to just "restart" the pod to have the new value for secret (envsubst does the same but at the cost of an init container).

PKizzle:
Who could be asked about the necessary additions to the harbor source code?
As using secrets is a basic requirement for any Kubernetes related project should this PR be accepted as temporary solution?

Vad1mo:
> Isn't this feature already implemented?
no, only on external DB, this one is for the internal DB.

PKizzle:
@Kajot-dev Yes, at least for the password. Would love to see usernames also being read from a secret but this PR definitely is no longer relevant.

PR #1334, opened 2022-10-28
Set default of updateStrategy matching to default accessMode

blackliner:
Because the default `accessMode` is `ReadWriteOnce` in the `value.yaml`, and since we know the combination with `RollingUpdate` [does not work](https://github.com/goharbor/harbor-helm/issues/408), I propose to change the default value.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -371,9 +371,10 @@ imagePullSecrets:
 
 # The update strategy for deployments with persistent volumes(jobservice, registry
 # and chartmuseum): "RollingUpdate" or "Recreate"
+# Set it as "RollingUpdate" when "RWM" for volumes is supported
 # Set it as "Recreate" when "RWM" for volumes isn't supported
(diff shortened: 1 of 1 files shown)


zyyw:
If users use cloud storage, we would prefer the `updateStrategy` defaults to "RollingUpdate" instead of "Recreate".
If users use pv as the backend storage, whether to use `updateStrategy` or `Recreate` really depends on the driver rather only the config of `RWO`/`RWM`.

Thanks for contributing to harbor-helm and closing it now.

blackliner:
@zyyw my issue is that the current default config is conflicting and not working anywhere. Or am I missing anything?

blackliner:
Specifically, the conflict comes from persistance being enabled by default and accessMode set to ReadWriteOnce

blackliner:
https://github.com/goharbor/harbor-helm/pull/470 introduced this

Issue #1348, opened 2022-11-21
Add option to disallow privilegeEscalation


relief-melone:
currently there is no option to edit the securityContext of containers inside of the various deployments of helm. It would be nice if this could be added. Would gladly add a PR if there are no objections to that option. Of course if there is no valid reason to enable privilegeEscalation the best thing would be to disable it in every container that could get started.

relief-melone:
https://github.com/goharbor/harbor-helm/pull/1349

I took the freedom to just create a PR. Feel free to reject or ask for any changes (The contributing guide was more on the acutal application so I am not sure if all here is correct)

zyyw:
thanks for contributing to Harbor-helm, unfortunately we have no plan to support `allowPrivilegeEscalation` so far.

relief-melone:
well the problem is you currently are. allowPrivilegeEscalation needs to be disabled explicitly and this prevents clusters which are safe by design and have policies in place to prevent pods that could "escalate" from running to being unable to run this chart

PR #1349, opened 2022-11-21
Feature/container security context

relief-melone:


justSem:
I was actually looking into these settings and saw your PR. I second this!
I do think just a securityContext block in the values file would be nice, because different clusters might have different requirements.
But being able to set things like capabilities and privilegeEscalations flags would be amazing!

relief-melone:
I actually had it in the values in my initial concept but ditched it because i saw the securityConcept of the pod was also hard coded. But I can commit a version with that in the values. Should I just make a suggestion or do you have something in mind?

justSem:
Well, what I was thinking was something like I see in more helm charts.
The helm chart offers the capability to the user to set the whole securityContext block (sometimes twice at pod/container level) but their default values are in line with the restricted security profile (where applicable).

So a default values could look like this:
```yaml
securityContext:
  runAsNonRoot: true
  capabilities:
    drop:
      - ALL
  allowPrivilegeEscalation: false
```

Setting that as default allows for default security, while keeping the flexibility for someone who wants to do custom stuff. (I.e.: our clusters enforce seccomp profiles which we now have to put in our deployments manually)

relief-melone:
this would leave security context for pod and container. Should we make both editable and do something like this in values

*values.yaml*
```yaml
securityContext:
  container:
    allowPrivilegeEscalation: false
   #...
  pod:
    runAsUser: 1000
    #...
```

or leave the pod securityContext hard coded as is?

justSem:
Hi there,

I apologise for the delayed response.
Looking at different Helm charts for different projects make it seem like most projects do what they deem to be the most fitting. There's nothing wrong with leaving the hard-coded values as-is imo, it's just that when someone wants to change that for some reason they'll have to edit the deployment(s) by hand.
But looking at the application as-is, I don't see a real use case for a need to edit the hardcoded values.

Kind Regards,
Sem

zyyw:
closing this pr per this comment:
* https://github.com/goharbor/harbor-helm/issues/1348#issuecomment-1333335716

Issue #1320, opened 2022-10-12
Harbor 2.6 - Jobservice cannot run replicated with RWO PVCs


Tonkari:
## Summary
Jobservice can only run as a single pod when the underlying kubernetes cluster does not support ReadWriteMany access mode.
## Environment
We are using this chart to deploy harbor on a kubernetes cluster that only supports ReadWriteOnce for persistent volumes. Currently we are running 3 replicas for the jobservice component. We are using the database JobLogger and S3 as storage backend for the registry.
## Problem
When installing or updating to chart version 1.10.0 (Harbor 2.6) only one jobservice pod will become available, while the others stay in "ContainerCreating" phase. This happens because they all try to mount the same new scan data export volume. As a workaround we can only reduce the number of replicas to 1, which we have been trying to avoid.
The chart does not allow for any configuration regarding the component. If persistence is enabled, a single scan data export volume will always be created and used for the deployment.
### Separate persistence option for scan data export
Without knowing, whether it's necessary to persist the scan data exports, a possible solution would be to add a separate option to disable persistence for this feature. Instead, exports could be stored in the pod's filesystem - long enough to download them - and disappear when the pod gets terminated. Since I have not really dug in the harbor code, I'm not sure if this is realistic.
### Store scan data exports in the database
The exports could be stored in the database. Same as above, I do not know if it's realistic, especially regarding the size of the generated csv files.
### External storage support for scan data export
Since the registry already has an S3 bucket to use for storage, the same could be done for the jobservice component. Whether the same bucket with a different path or a different bucket should be used, can be discussed.
### Use statefulSet for jobservice
The jobservice deployment could be converted to a statefulSet, and a volumeClaimTemplate could be created for the scan data exports. For the file JobLogger the persistentVolumeClaim could remain the same.
### Update docs
If none of the above solutions are desirable, this incompatibility should be mentioned in the docs.

r-ising:
We have the same issue after upgrading to harbor 2.6

niklasweimann:
+1

janosmiko:
+1

Duke1616:
+1

blufor:
I can confirm this is a valid bug

maxdanilov:
Struggling with the same issue, I went for an approach of choosing HA over scan data export, by manually fixing to enable HA for jobservice would be to switch in the resulting manifests from:

```yml
      - name: job-scandata-exports
        persistentVolumeClaim:
[...]

PR #1354, opened 2022-12-08
Add an ability to disable PVC for scandata

eugenberend:
To deal with https://github.com/goharbor/harbor-helm/issues/1320, we should add ability to completely disable creation and usage of ScanData PVC.


zyyw:
Since we have this PR:
* https://github.com/goharbor/harbor-helm/pull/1378

The current PR is no need any more! Closing it now.

PR #1372, opened 2023-01-18
Support external secrets for service tokens and the xsrf key

darend:
Add support for using external/existing secrets for the following keys:

- registry.secret
- jobservice.secret
- core.secret
- core.xsrfKey


darend:
Replaced with https://github.com/goharbor/harbor-helm/pull/1373

PR #1389, opened 2023-02-09
remove default tls keys

chaospuppy:
...and expand options for supplying/autogenerating them.

Having the [public token signing keys](https://github.com/goharbor/harbor-helm/tree/master/cert) used by default essentially makes the Harbor Helm chart insecure by default.  It is easy to miss that if you do not specify a preexisting `core.secretName` value, that a publicly disclosed key will be used to sign tokens issued from the Harbor API.

This PR will remove those certificates, adds a new `core.tokenKey` value where a PEM-encoded RSA key can be specified, and if neither `core.tokenKey` or `core.secretName` are provided, the helm template for the core secret will autogenerate a unique RSA key.

Additionally, this PR removes the `tls.crt` key from secrets, as it is not used anywhere, and also removes a portion of the `harbor-exporter` secret, which also was not used. 

Signed-off-by: Tim Seagren <seagren.tim@solute.us>

PR #1403, opened 2023-02-20
add docs, values, and template updates to remove default tls key and …

zyyw:
…autogenerate if no key/cert pair is supplied (#1389)


wy65701436 (approved):
lgtm

ywk253100 (approved):

PR #1423, opened 2023-02-27
feat: add init containers

kimxogus:
Add init containers to each deployment components


zyyw:
@kimxogus could you please justify why we need this init containers?

kimxogus:
@zyyw As harbor-registry component doesn't check if the pod has proper iam permission(not like https://github.com/distribution/distribution), I'd like to prevent pod from running if pod doesn't have proper iam permission.

As I have to use kiam or kube2iam to use harbor-registry, there could be mistakes like assigning wrong node groups which don't have permission to assume kiam/kube2iam role.

rwd5213:
Hi, we actually need this for Trivy init container. when in offline environments we need an init container that downloads the vulnerability data base to the trivy container

Vad1mo:
IMO, Trivy can download the DB itself? Why are you using the init container?

rwd5213:
Because of environment where it cannot download the database itself, we use an init container to pull the database from S3 for example. I am not aware of a way to directly tell Trivy where to download the data base from

Vad1mo:
I don't find this to be a bad idea, there are a many uses cases that I can think of that would simplify the user of harbor with an init container. 

@kimxogus @rwd5213 can you rebase and resolve the conflicts, I would like to get those changes in.

Vad1mo (approved):

rwd5213:
@kimxogus I realized this is missing the init container in the in the https://github.com/goharbor/harbor-helm/blob/main/templates/trivy/trivy-sts.yaml, any chance you can add it there as well?

kimxogus:
Merged current main and also applied to trivy statefulset.

MaximilianB134:
Hi, I just wanted to state that we also need init containers, if you guys might want another justification for it. Our storage class always creates PVCs in a way that only root can write to it and with an init container we could automate/streamline the setup process so that harbor works directly and no manual intervention is needed.

MinerYang (changes requested):
Shall we also add this to database sts?

kimxogus:
@MinerYang I've just added init containers for database and redis.

zyyw (approved):

Vad1mo:
@kimxogus -> Merging can be performed automatically once the requested changes are addressed.

MinerYang (approved):

Issue #1358, opened 2022-12-21
Chart does not allow Postgresql secret instead of password


redbelow:
https://github.com/goharbor/harbor-helm/blob/185354a01bc00fb9db5943dfd331128fbcd149d8/templates/_helpers.tpl#L138

Hi, the current chart (1.11.0) does not allow secret instead of postgres password to be used (for either internal or external db). Attempting to deploy without the password for postgres will cause service failure.

I think fixing this will require changes to the harbor image itself as well as the chart.

torsten-online:
Yes, I have the same problem. The existing secret is not used for the password of notary-server and notary-signer database, the default value "password" is used.

figaw:
Ideally it would be possible to use an `existingSecret`, to configure both 

`NOTARY_SERVER_DB_URL` and `NOTARY_SIGNER_DB_URL` inside the pods.

---

If you need it "right now" you can specify the password under `database.external.password` in the `values.yaml`-file. 

The helm chart uses this value in the `_helpers.tpl` to generate the db_url:

`_helpers.tpl`: 
```helm
{{- define "harbor.database.notaryServer" -}}
postgres://{{ template "harbor.database.username" . }}:{{ template "harbor.database.escapedRawPassword" . }}@{{ template "harbor.database.host" . }}:{{ template "harbor.database.port" . }}/{{ template "harbor.database.notaryServerDatabase" . }}?sslmode={{ template "harbor.database.sslmode" . }}
{{- end -}}
```

where `harbor.database.escapedRawPassword` comes from `Values.database.external.password`:

```helm
{{- define "harbor.database.rawPassword" -}}
  {{- if eq .Values.database.type "internal" -}}
    {{- .Values.database.internal.password -}}
  {{- else -}}
    {{- .Values.database.external.password -}}
  {{- end -}}
{{- end -}}

{{- define "harbor.database.escapedRawPassword" -}}
  {{- include "harbor.database.rawPassword" . | urlquery | replace "+" "%20" -}}
{{- end -}}
```

The `harbor.database.notaryServer` is used by the `templates/notary/notary-secret.yaml` in both:

```yaml
  server.json: {{ tpl (.Files.Get "conf/notary-server.json") . | b64enc }}
  ...
  NOTARY_SERVER_DB_URL: {{ include "harbor.database.notaryServer" . | b64enc }}
```

And the `conf/notary-server.json` (inside the secret):

```
{ 
  ...
  "storage": {
    "backend": "postgres",
[...]

PR #1425, opened 2023-02-27
allow using existing secret for notary database URL

ghudiczius:
- allow using existing secret for setting the database URL for notary when using an external database
- solves #1358
- did not touch the server.json and signer.json templates as the incorrect database URL in those will be overwritten
- added the `NOTARY_SERVER_STORAGE_DB_URL` and `NOTARY_SIGNER_STORAGE_DB_URL` environment variables to the notary-server and notary-signer containers respectively which use the existing secret to inject the database URL into the containers
  - these variables follow the naming scheme to overwrite configuration values (see https://github.com/notaryproject/notary/blob/master/docs/running_a_service.md#overriding-configuration-file-parameters-using-environment-variables)

### DIFF
--- a/templates/notary/notary-secret.yaml
+++ b/templates/notary/notary-secret.yaml
@@ -17,6 +17,8 @@ data:
   {{- end }}
   server.json: {{ tpl (.Files.Get "conf/notary-server.json") . | b64enc }}
   signer.json: {{ tpl (.Files.Get "conf/notary-signer.json") . | b64enc }}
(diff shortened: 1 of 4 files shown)


zyyw (approved):

zyyw:
@ghudiczius , thank you for contributing to harbor-helm. We do appreciate your effort.

As notary and chartmuseum are going to be deprecated and removed in the near future. We may not accept PR regarding notary chartmuseum.

PR #1455, opened 2023-03-31
feat: Allow specifying strategy for deployments

hauke-d:
Fix: #1469

Proposing to add a `strategy` field to components with a scalable `Deployment`. In case of many replicas one might want to have tighter control of how how updates are rolled out.

Furthermore, we have the following use-case: We are running a HA deployment with 2 registry replicas without persistence and an `updateStrategy.type` of `Recreate` (which we need for the stateful components). Even though our registry replicas are stateless without volumes, both of them are taken down at the same time when we apply updates due to the `Recreate` option. So we have downtime whenever there are changes, although we would like to provide HA service.  

This change would allow us to control the update strategy for each component individually, if necessary.


LiuShuaiyi:
So there's already a "global" upgradeStrategy value that will apply to both registry and jobservice https://github.com/goharbor/harbor-helm/blob/master/values.yaml#L351, but in your case you want to set different upgradeStrategy values for registry and jobservice, right? 

I think this is a legit use case, but I would suggest the followings:
1. There don't seem to be needs for components other than jobservice and registry to set upgradeStrategy
2. We can remove the "global" upgradeStrategy value setting once it can be set individually

hauke-d:
Hi @LiuShuaiyi 

Thanks, I just included it for all components as it doesn't hurt to be able to modify it if use-cases should come up, but I can of course remove it if you prefer to keep the options limited to what makes sense at the moment. 

I considered splitting the global `updateStrategy` up and remove it, but did not do it initially as it would be a breaking change. If you agree it's worth it, I'd be ready to modify this PR accordingly.

Issue #1433, opened 2023-03-03
Make the Repo to be indicated as the "Official" repo for Harbor


Vad1mo:
<img width="1265" alt="image" src="https://user-images.githubusercontent.com/1492007/222788144-0dcce463-910f-4855-920f-a0ca78c3ff38.png">

In Artifact Hub, the official status means that the publisher owns the software deployed by a package. If we consider the example of a [chart used to install Consul](https://artifacthub.io/packages/helm/hashicorp/consul), to obtain the official status the publisher should be the owner of the Consul software (HashiCorp in this case), not just the chart.

The official status can be granted at the repository or package level. When it is granted for a repository, all packages available on it will display the official badge, so all packages in the repository must be official. If only some of the packages in your repository are official, please list them in the Official packages field when submitting the official status request.

Before applying for this status, please make sure your repository complies with the following requirements:

The repository has already obtained the [Verified publisher](https://artifacthub.io/docs/topics/repositories/#verified-publisher) status.
The user requesting the status is the publisher of the repository in Artifact Hub, or belongs to the organization publishing it.
All official packages available in the repository provide a README.md file with some documentation that can be displayed on Artifact Hub.
Once you have verified that the requirements are met, please file an issue [using this template](https://github.com/artifacthub/hub/issues/new?assignees=&labels=official+status+request&template=official-status-request.md&title=%5BOFFICIAL%5D+Your+repository+or+project+name) to apply.

See:
https://artifacthub.io/docs/topics/repositories/#official-status

OrlinVasilev:
https://github.com/artifacthub/hub/issues/2970

OrlinVasilev:
![image](https://user-images.githubusercontent.com/7236111/230599869-fa135982-b89c-43ba-8fe0-8e4d97a5ab0c.png)
Done :)

Issue #1432, opened 2023-03-03
Make this repo a Verified publisher on artifacthub.io


Vad1mo:
Repositories and the packages they provide can display a special label named Verified publisher. This label indicates that the repository publisher owns or has control over the repository. Users may rely on it to decide if they want to use a given package or not.

Publishers can be verified through the [artifacthub-repo.yml](https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-repo.yml) repository metadata file. In the repositories tab in the Artifact Hub control panel, the repository identifier is exposed on each repository’s card (ID). To proceed with the verification, an artifacthub-repo.yml metadata file must be added to the repository including that ID in the repositoryID field. The next time the repository is processed, the verification will be checked and the flag will be enabled if it succeeds.

Please note that the artifacthub-repo.yml metadata file must be located at the repository URL’s path. In Helm repositories, for example, this means it must be located at the same level of the chart repository index.yaml file, and it must be served from the chart repository HTTP server as well.

The verified publisher flag won’t be set until the next time the repository is processed. Please keep in mind that the repository won’t be processed if it hasn’t changed since the last time it was processed. Depending on the repository kind, this is checked in a different way. For Helm http based repositories, we consider it has changed if the index.yaml file changes (the generated field is ignored when performing this check). For git based repositories, it does when the hash of the last commit in the branch you set up changes.

https://artifacthub.io/docs/topics/repositories/#verified-publisher

OrlinVasilev:
![image](https://user-images.githubusercontent.com/7236111/230595804-77c741a2-f15e-47f4-962d-2d13ba01d139.png)

PR #1459, opened 2023-04-06
Add artifacthub-repo.yml to be able to verify publisher

OrlinVasilev:
fixes https://github.com/goharbor/harbor-helm/issues/1433 and https://github.com/goharbor/harbor-helm/issues/1432

### DIFF
--- a/artifacthub-repo.yml
+++ b/artifacthub-repo.yml
@@ -1,0 +1,25 @@
+# Artifact Hub repository metadata file
+#
+# Some settings like the verified publisher flag or the ignored packages won't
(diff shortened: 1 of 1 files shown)


zyyw (approved):

zyyw (changes requested):
the `artifacthub-repo.yml` should be added within the branch of `chart-repository`:
```
Please note that the artifacthub-repo.yml metadata file must be located at the repository URL’s path. In Helm repositories, for example, this means it must be located at the same level of the chart repository index.yaml file, and it must be served from the chart repository HTTP server as well.
```

PR #1470, opened 2023-04-14
feat: update database-ss initContainers for photon 4.0

zyyw:


values.yaml
```diff
@@ -275,8 +275,8 @@ proxy:
 # If expose the service via "ingress", the Nginx will not be used
 nginx:
   image:
-    repository: goharbor/nginx-photon
-    tag: v1.10.17
+    repository: yangjiao0817/nginx-photon
```
Vad1mo, line 278: is it  correct to use prefix yangjiao0817?

Vad1mo (changes requested):
there are some copy and past errors

values.yaml
zyyw: that's for testing purpose. It has been updated now. Thanks

MinerYang (approved):

ywk253100 (approved):

Vad1mo (approved):

PR #1474, opened 2023-04-17
feat: add config of reaper.max_update_hours & reaper.max_dangling_hours

zyyw:
add jobservice config of this: https://github.com/goharbor/harbor/blob/main/make/photon/prepare/templates/jobservice/config.yml.jinja#L55-L59


values.yaml
```diff
@@ -515,6 +515,11 @@ jobservice:
   notification:
     webhook_job_max_retry: 3
     webhook_job_http_client_timeout: 3 # in seconds
+  reaper:
```
MinerYang, line 518: It is more like a env variable that have not been exposed to the harbor.yml 
https://github.com/goharbor/harbor/blob/main/make/harbor.yml.tmpl
Do we need it configurable in harbor-helm?

stonezdj: It is ok to keep it as it was

stonezdj (approved):

values.yaml
zyyw: it is configured in the `harbor/common/config/jobservice/config.yml`. If you deploy a Harbor v2.8.0, you will find it there. Therefore, we add it in `jobservice-cm.yaml`, instead of `jobservice-cm-env.yaml`.

MinerYang (approved):

PR #1490, opened 2023-04-28
Add external proxy option

CloudlightCoUk:
The Harbor architecture describes a Proxy layer, but the Helm chart doesn't allow for that layer to be provided by anything other than nginx, with either an nginx object being created if the expose type is "ingress", or nginx itself being deployed if set to anything else.

I would like to extend the functionality to be able to 'bring my own proxy', in my case I want to use Contour with HTTPproxy objects. Rather than implement support for this in Harbor, it is easier simply to implement and expose option of 'external' where the chart can assume the proxy deployment is handled elsewhere. In this case no nginx deployment or object creation is triggered by the chart.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -8,7 +8,7 @@ keywords:
 - registry
 - harbor
 home: https://goharbor.io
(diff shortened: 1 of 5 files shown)


CloudlightCoUk:
Made a miss, will resubmit

PR #1494, opened 2023-05-05
feat:Add external proxy option

CloudlightCoUk:
The Harbor architecture describes a Proxy layer, but the Helm chart doesn't allow for that layer to be provided by anything other than nginx, with either an nginx object being created if the expose type is "ingress", or nginx itself being deployed if set to anything else.

I would like to extend the functionality to be able to 'bring my own proxy', in my case I want to use Contour with HTTPproxy objects. Rather than implement support for this in Harbor, it is easier simply to implement and expose option of 'external' where the chart can assume the proxy deployment is handled elsewhere. In this case no nginx deployment or object creation is triggered by the chart.

### DIFF
--- a/templates/nginx/configmap-http.yaml
+++ b/templates/nginx/configmap-http.yaml
@@ -1,4 +1,4 @@
-{{- if and (ne .Values.expose.type "ingress") (not .Values.expose.tls.enabled) }}
+{{- if and (ne .Values.expose.type "ingress") (ne .Values.expose.type "external") (not .Values.expose.tls.enabled) }}
 {{- $scheme := (include "harbor.component.scheme" .) -}}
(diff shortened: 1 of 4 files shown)


Vad1mo:
same as https://github.com/goharbor/harbor-helm/pull/1073

CloudlightCoUk:
I saw https://github.com/goharbor/harbor-helm/pull/1073, but it hasn't moved in a year, so thought I'd share what I have done.

I'm not massively opinionated on the implementation, but do feel strongly that as Harbor is a layered architecture, we should have flexibility on the proxy we use, just as we do the storage/db/cache, we use. Especially as nginx is becoming less popular for proxy/ingress.

PR #1497, opened 2023-05-08
Fix the nginx's containerPort

kariya-mitsuru:
It is better to limit the nginx's `containerPort` and `port` to only those that are enabled.

`8443` is valid only if `.Values.expose.tls.enabled` is `true`.
`4443` is valid only if `.Values.expose.tls.enabled` and `.Values.notary.enabled` are both `true`.

### DIFF
--- a/templates/nginx/deployment.yaml
+++ b/templates/nginx/deployment.yaml
@@ -79,8 +79,9 @@ spec:
 {{- end }}
         ports:
         - containerPort: 8080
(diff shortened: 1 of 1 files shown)


zyyw:
Hi @kariya-mitsuru , as Notary is now removed from master branch, could you please rebase master into your PR branch? Thanks

zyyw (approved):

Vad1mo (approved):

Issue #1181, opened 2022-04-26
Add optional PodDisruptionBudget to chart


slushysnowman:
Currently, there is no option to create PodDisruptionBudget in the chart. It would be very handy to have the option to do this from within the chart, so that users who require a PodDisruptionBudget don't have to maintain this separately.

zyyw:
Since the PDB is a relatively independent resource. We would suggest you manage PBD separately instead of incorporating with harbor-helm.

robinkb:
I disagree with this conclusion. I consume a lot of Helm charts in my environments, and most Helm charts which can setup highly-available applications include PDBs in the chart. Having to configure all of the matchLabels and such is a hassle that I would expect the Helm chart to do for me.

Having said that, I do think that the proposed PR (#1182) was not very elegant. My favorite implementation that I've seen so far is in the [chart for oauth2-proxy](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/poddisruptionbudget.yaml). The PDBs are enabled by default, but only when the replicaCount is > 1.

PodDisruptionBudgets are essential to ensure that applications don't go down during normal cluster operations, like rolling restarts of nodes.

PR #1509, opened 2023-05-17
Install PodDistruptionBudgets when replica greater than 1

rgarcia89:
Currently there is no pdb deployed for components were the replica is defined greater than 1. This can lead to an outage of the components during a cluster update. The added pdb definitions will be added automatically once the replica size is greater than 1 and ensure that at least one replica is ready before the pod will be drained.


rgarcia89:
@Vad1mo ready for review

Vad1mo:
I like it. It would be beneficial to make it optional:

- [ ]  add enable/disable flag, on each component an  
- [ ] Allow to set
Min amount

rgarcia89:
Is setting the min amount really necessary? Currently at least one pod is necessary in order for the component to fulfill its tasks. Therefore I think we should be good with the fix value.

Vad1mo (approved):

PeopleOfPlay:
Thanks for the implementation, it's really cool! 

I quite agree with @Vad1mo and I understand your point of view, but harcoding values is never very good, some users may have use cases that push them to change this number. (If, for example, the harbor has a lot of simultaneous users, perhaps two replicas of a certain service is the bare minimum required to maintain the desired quality of service when operating on Kubernetes nodes.) 

In issue #1181 robinkb proposed a rather elegant solution from the [oauth2-proxy chart](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/templates/poddisruptionbudget.yaml).

I think it would be better to set the default value to 1 and leave it up to the user to choose what best suits their needs. (ref [helm doc](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function))

It might also be cool to have the same logic with maxUnavailable fields. (ref #1182)

MarcHenriot:
@rgarcia89 I have pushed a version of the above changes in my fork, can we merge my version to yours ?

rgarcia89:
> I forgot to sign my commits …. What do we do ?

Yeah I just signed them one by one. Should be fine now

rgarcia89:
@Vad1mo I think this PR is ready to be merged

MarcHenriot:
There is typo in all pdb, I will push a fix to your fork.

It misses `matchLabels` in the `selector` section.

rgarcia89:
@Vad1mo any idea when this can be merged?

OrlinVasilev:
@zyyw  can you please review that PR thank you!

rgarcia89:
@MinerYang @Kajot-dev maybe you can have a look here too

slushysnowman:
Would be great to get this in - at the moment we're manually adding it for our Harbor setup

Kajot-dev:
That would be great to have! We're adding these besides the helm chart currently

balonik:
@zyyw please review and approve this MR. It is pending just single approval.

OrlinVasilev:
@Vad1mo I think that's good to go

OrlinVasilev (approved):

phantasm66:
It looks like the only failing workflow integration test is due to an outdated workflow action (`deprecated version of `actions/upload-artifact: v2``). Can that be addressed and this feature be merged? 🙏

rgarcia89:
All cleared and ready to be merged

amar-opsverse (approved):

quercus-carsten:
Hello, would it be possible to have this beneficial improvement reviewed? I'm not certain, but it appears that @zyyw may not have been active recently. Could someone else possibly assist?

Kajot-dev:
All the requirements are met. Can somebody proceed with the merge before we hit 2.13.0?

OrlinVasilev (approved):
LGTM, it's about time to merge this one!

rgarcia89:
I am really wondering to see how many people it needs to get something merged. Especially given the fact that there are already 3 approvals :sweat_smile:

bupd (approved):
lgtm

quercus-carsten:
User @zyyw was requested to review, but it appears that the user has been inactive since September 2024.
There have been no contributions since then.

Is there anyone in the Harbor project who can help out?

MinerYang (changes requested):
Hi @rgarcia89 ,
By setting PDB  minAvailable: 1 for each of the pod will have these impacts:
-  **Would create a direct conflict with resource management practices**. For example Kubernetes cannot voluntarily move or evict any of the pods from their current nodes. 
- And a PDB violation can also slow down the rolling upgrade process if new nodes aren't available to place replacement pods on, leading to increased deployment times
- PDBs are only effective for pod running with more than one replica. **A PDB for a single-replica application with minAvailable: 1 will block all voluntary disruptions, preventing node maintenance.**
  
PDB sacrifices cluster flexibility, efficiency, and maintainability. I personally do not see it as a necessarily configuration for pod and as an overhead operation for most of the Harbor deploy scenario.

So please clarify your using scenario and necessity justification  and better to make it optional.

Best,
Miner

Kajot-dev:
> Hi @rgarcia89 , By setting PDB minAvailable: 1 for each of the pod will have these impacts:
> 
> * **Would create a direct conflict with resource management practices**. For example Kubernetes cannot voluntarily move or evict any of the pods from their current nodes.
> * And a PDB violation can also slow down the rolling upgrade process if new nodes aren't available to place replacement pods on, leading to increased deployment times
> * PDBs are only effective for pod running with more than one replica. **A PDB for a single-replica application with minAvailable: 1 will block all voluntary disruptions, preventing node maintenance.**
> 
> PDB sacrifices cluster flexibility, efficiency, and maintainability. I personally do not see it as a necessarily configuration for pod and as an overhead operation for most of the Harbor deploy scenario.
> 
> So please clarify your using scenario and necessity justification and better to make it optional.
> 
> Best, Miner

I must disagree.

The whole point of PDB is to prevent k8s from evicting all your pods at once and causing a downtime for your service.

> For example Kubernetes cannot voluntarily move or evict any of the pods from their current nodes.

That would be the case if allowed disruptions was `0` (eg. minAvailable when replicas is 1) Of course it is possible to create bad configuration, via anti-affinity or nodeSelector, but that's like saying we should not use `nodeSelector` since you can specify an invalid one.

Perhaps pdb could be made more configurable or made opt-in, but these are an essential part of any ha deployment. I've been using `pdb` with Harbor on multiple production instances for a couple of years and these have prevented many outages during maintenance, cluster upgrade, etc.

I hope that we can find a solution that works for everyone.

MinerYang (approved):
lgtm

ricardojdsilva87:
Hello,
This feature was not added to any release yet, it's only on the main branch, the `1.18.3` tag was released around 1 hour ago.
Could you please check?
I've also noticed that these [2 lines ](https://github.com/goharbor/harbor-helm/blame/main/values.yaml#L688-L689)on the `values.yaml` file, are duplicated, but were here before this change was introduced. Most likely the lines were copy pasted across the different services and the already existing values were not noticed.
Thanks!

afreyermuth98:
agreed with @ricardojdsilva87 , do you have an ETA for the release ?

flaviomoringa:
Such a simple thing... please merge this asap...  make sure others don't have to suffer what happened with us because of not having this in place.

PR #1514, opened 2023-05-28
Fix incorrect comment on `persistance.imageChartStorage.gcs.existingSecret` in the values.yaml

arjungandhi:
The comment specifies that the key for the secret should be `gcs-key.json` 

However on [line 291](https://github.com/goharbor/harbor-helm/blob/82a6f3649ea54464b70eeea3ce2abd5499785e93/templates/registry/registry-dpl.yaml#L291) the key existing secrets use is actually `GCS_KEY_DATA`


ywk253100 (approved):

OrlinVasilev:
@arjungandhi congrats on merging your first PR here :))

PR #1522, opened 2023-06-14
Allow customization of annotations on services exposing metrics

afreyermuth98:
I want to annotate the services that expose metrics to scrape them.


values.yaml
```diff
@@ -788,6 +792,7 @@ redis:
     existingSecret: ""
   ## Additional deployment annotations
   podAnnotations: {}
+  svcAnnotations: {}
```
zyyw, line 795: why here it is `svcAnnotations` while the others are `serviceAnnotations`?

afreyermuth98: @zyyw Oh yeah my bad. I firstly did `svcAnnotations` and then I realized that it was already done on core and notary component as `serviceAnnotations` so I've adapted the syntax and missed one ^^'

zyyw:
Thanks for contributing to harbor-helm.

Please also follow this instruction:
* https://github.com/goharbor/harbor-helm/pull/1522/checks?check_run_id=14280834816
to fix the `DCO` failure issue in CI.

afreyermuth98:
Up :)

Vad1mo (approved):

PR #1547, opened 2023-07-19
Add log format to registry config

Nefelim4ag:


zyyw:
Hi @Nefelim4ag could you please resolve the `DCO`?
BTW, this feature needs to be configurable in harbor.yml file under [harbor oss project](https://github.com/goharbor/harbor) first, then we can consider to pick up this configurations in harbor-helm.

Nefelim4ag:
This configuration only affects the registry, which initially is a fork of the docker distribution.
AFAIK, as I saw in the harbor code, there is a custom in-house logger everywhere and no support for JSON logging.
So, I'm not sure what needs to be changed in [harbor oss project](https://github.com/goharbor/harbor)

Vad1mo (approved):

PR #1550, opened 2023-07-21
Fix: Avoid recreating random secrets

### LINKED ISSUE #1549: Helm chart recreate secrets on every deploy
On every deployment, the helm chart is recreating the secrets and redeploying the core, registry and jobservices pods as we are using `randAlphaNum` function. For instance:

https://github.com/goharbor/harbor-helm/blob/77de1dac06f9b4a9aff8e0fc6032eca001169e7e/templates/core/core-secret.yaml#L12
https://github.com/goharbor/harbor-helm/blob/77de1dac06f9b4a9aff8e0fc6032eca001169e7e/templates/registry/registry-secret.yaml#L9

xoanmi:
Avoid recreating secrets with dynamic/random content on every helm update. This causes the `core`, `registry` and `jobservice` services to be redeployed in each update.

Related to https://cloud-native.slack.com/archives/CC1E09J6S/p1689925408303219

Close #1549

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -528,22 +528,26 @@ app: "{{ template "harbor.name" . }}"
 {{- $secret := (lookup "v1" "Secret" .Release.Namespace (include "harbor.core" .) ) -}}
 {{- if $secret -}}
 {{/* Reusing existing secret data */}}
-secret: {{ index $secret "data" "secret" }}
 CSRF_KEY: {{ index $secret "data" "CSRF_KEY" }}
+HARBOR_ADMIN_PASSWORD: {{ index $secret "data" "HARBOR_ADMIN_PASSWORD" }}
+POSTGRESQL_PASSWORD: {{ index $secret "data" "POSTGRESQL_PASSWORD" }}
+REGISTRY_CREDENTIAL_PASSWORD: {{ index $secret "data" "REGISTRY_CREDENTIAL_PASSWORD" }}
+secret: {{ index $secret "data" "secret" }}
+secretKey: {{ index $secret "data" "secretKey" }}
 tls.key: {{ index $secret "data" "tls.key" }}
 tls.crt: {{ index $secret "data" "tls.crt" }}
 {{- else -}}
 {{/* 
   Generate new data
 */}}
+{{- if not .Values.existingSecretSecretKey }}
+secretKey: {{ .Values.secretKey | b64enc | quote }}
+{{- end }}
 secret: {{ .Values.core.secret | default (randAlphaNum 16) | b64enc | quote }}
-CSRF_KEY: {{ .Values.core.xsrfKey | default (randAlphaNum 32) | b64enc | quote }}
+{{- if not .Values.core.secretName }}
 {{- $ca := genCA "harbor-token-ca" 365 }}
 tls.key: {{ .Values.core.tokenKey | default $ca.Key | b64enc | quote }}
 tls.crt: {{ .Values.core.tokenCert | default $ca.Cert | b64enc | quote }}
-{{- end }}
-{{- if not .Values.existingSecretSecretKey }}
-secretKey: {{ .Values.secretKey | b64enc | quote }}
 {{- end }}
 {{- if not .Values.existingSecretAdminPassword }}
 HARBOR_ADMIN_PASSWORD: {{ .Values.harborAdminPassword | b64enc | quote }}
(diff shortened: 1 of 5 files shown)


Kajot-dev:
Actually, I think that `tls.key` and `tls.cert` can be defined using existing Secret via `core.secretName` which is something that is separate from the secret created by the chart (`harbor.core` template). This PR removes it and assumes that all the required values are already present it the secret if it exists.

While this PR will work for new deployment it removes defining certificate for token signing from external secret (you can use CertManager to do just that) - it **will** break the installation for those who already use `core.secretName` by not using it and inserting `nil` values into the `-core` secret (since `-core` secret exists and does not have `tls.cert` and `tls.key`)

The idea is good though - but it needs to preserve the `core.secretName` feature and it needs to check for presence of each individual value inside the secret.

xoanmi:
> Actually, I think that `tls.key` and `tls.cert` can be defined using existing Secret via `core.secretName` which is something that is separate from the secret created by the chart (`harbor.core` template). This PR removes it and assumes that all the required values are already present it the secret if it exists.
> 
> While this PR will work for new deployment it removes defining certificate for token signing from external secret (you can use CertManager to do just that) - it **will** break the installation for those who already use `core.secretName` by not using it and inserting `nil` values into the `-core` secret (since `-core` secret exists and does not have `tls.cert` and `tls.key`)
> 
> The idea is good though - but it needs to preserve the `core.secretName` feature and it needs to check for presence of each individual value inside the secret.

Hi @Kajot-dev it should be solved. Could you review it again please?

templates/_helpers.tpl
```diff
@@ -521,4 +521,140 @@ app: "{{ template "harbor.name" . }}"
...
+{{- if not .Values.database.external.existingSecret }}
+POSTGRESQL_PASSWORD: {{ template "harbor.database.encryptedPassword" . }}
+{{- end }}
+{{- if not .Values.registry.credentials.existingSecret }}
+REGISTRY_CREDENTIAL_PASSWORD: {{ .Values.registry.credentials.password | b64enc | quote }}
+{{- end }}
+CSRF_KEY: {{ .Values.core.xsrfKey | default (randAlphaNum 32) | b64enc | quote }}
+{{- if .Values.core.configureUserSettings }}
+CONFIG_OVERWRITE_JSON: {{ .Values.core.configureUserSettings | b64enc | quote }}
+{{- end }}
```
Kajot-dev, line 564: This is great but it seems that `CONFIG_OVERWRITE_JSON` is present only when generating new data

templates/_helpers.tpl
```diff
@@ -521,4 +521,140 @@ app: "{{ template "harbor.name" . }}"
...
+{{- end -}}
+{{- template "harbor.traceJaegerPassword" . }}
+{{- end -}}
+
+{{/* Harbor JobService Secret generator */}}
+{{- define "harbor.jobservice.secret" -}}
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace (include "harbor.jobservice" .) ) -}}
+{{- if $secret -}}
+{{/* Reusing existing secret data */}}
```
Kajot-dev, line 573: Why is there no `REGISTRY_CREDENTIAL_PASSWORD`?
```
{{- if not .Values.registry.credentials.existingSecret }}
REGISTRY_CREDENTIAL_PASSWORD: {{ .Values.registry.credentials.password | b64enc | quote }}
{{- end }}
```

Kajot-dev: (when reusing existing data)

Kajot-dev (changes requested):
First, you did a lot of work, thanks!!

Nice changes, but some of them apply only when generating new data.
Also there is much code duplication like this:
```
{{- $secret := lookup ... }}
{(- if $secret }}
old data..
{{- else }}
new data
{{- end }}
```
Also not all secret values are reused - for example if user won't put S3 secrets in external secret it will not use lookup for them

If you want to you can avoid code duplications by leveraging the fact that when secret is not present helm returns an empty object. Instead of having two blocks for reusing data and generation of the new data which share most of the code you can do something like this:
```helm
{{/* template definition */}}
{{- define "harbor.keyExtractor" -}}
{{- if hasKey .data .key }}
{{- $val := (index .data .key) }}
{{- if kindIs "string" $val }}
{{ .key }}: {{ $val | quote }}
{{- else }}
{{ .key }}: {{ $val }}
{{- end }}
{{- else }}
{{ .default }}
{{- end }}
{{- end -}}

{{/* and then */}}
{{- $secret := lookup ... }}
{{- template "harbor.keyExtractor" (dict "key" "SOME_PASSWORD_KEY" "data" $secret.data "default" "DEFAULT_VALUE" ) }}

templates/_helpers.tpl
xoanmi: Is there, isn't it? https://github.com/goharbor/harbor-helm/pull/1550/files#diff-87d68c754766af8e2e930e653be7e4b75fa0c8bdb187cb1bec293f265d9159ffR538-R540

templates/_helpers.tpl
xoanmi: Added!

templates/_helpers.tpl
Kajot-dev: Yes, my bad

Kajot-dev (approved):
looks good to me
Nice work!

Kajot-dev:
Could we get this reviewed and merged please?

Kajot-dev:
fix: #1609

zyyw:
Thanks for contributing to harbor-helm! 
Closing this PR now as it was fixed in PR:
* https://github.com/goharbor/harbor-helm/pull/1612

PR #1573, opened 2023-08-17
:wrench: Redis replica

afreyermuth98:
I want to set the replica count of redis


zyyw:
We currently do not plan to expose replicas for statefulset resources, such as `db` & `redis`. The pods of statefulset have their own pv and data-sync among different pv(s) associated with corresponding pod may require additional efforts.

PR #1582, opened 2023-08-30
[Feature] Harbor branding

Kajot-dev:
This PR enables Harbor branding via the helm chart by mounting `setting.json` file inside portal deployment.
Additionally, it can mount image files from `ConfigMaps` to be used as logo and/or login background (since configmaps can store small binary files)

### DIFF
--- a/templates/portal/configmap.yaml
+++ b/templates/portal/configmap.yaml
@@ -65,3 +65,8 @@ data:
             }
         }
     }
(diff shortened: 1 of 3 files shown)


zyyw:
Hi @Kajot-dev , we currently have no plan for this. Instead of configuring harbor-helm for portal attributes, you may try to build a customized harbor-portal image and use that customized harbor-portal when you deploying harbor-helm.

PR #1595, opened 2023-09-04
Add apiVersion and kind to sts volume claim templates

LinAnt:


zyyw:
We may also need to update the same part for `templates/database/database-ss.yaml` as well.

zyyw:
@LinAnt thank you for contributing to harbor-helm! Could you please update the same part for `templates/database/database-ss.yaml` as well?

Issue #1599, opened 2023-09-07
when uninstall harbor, migration-job should be delete


kycheng:
After uninstalling harbor, it is found that there is a job left that has not been deleted (deployed in k8s).

```
cluster-console> kgpo  | grep v223-harbor1                                                                                                                                                                            
Warning: Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
v223-harbor1-fg6hv-harbor-core-migration-job-6n4fw    0/1     Error       0               61m
v223-harbor1-fg6hv-harbor-core-migration-job-ddctg    0/1     Error       0               61m
v223-harbor1-fg6hv-harbor-core-migration-job-kdw5m    0/1     Completed   0               59m
v223-harbor1-fg6hv-harbor-core-migration-job-qw8z7    0/1     Error       0               60m
cluster-console> kg job v223-harbor1-fg6hv-harbor-core-migration-job
Warning: Use tokens from the TokenRequest API or manually created secret-based tokens instead of auto-generated secret-based tokens.
NAME                                           COMPLETIONS   DURATION   AGE
v223-harbor1-fg6hv-harbor-core-migration-job   1/1           2m45s      61m
```

The cause of the problem should be that the helm hook mechanism is used, causing the release to no longer manage the job.

https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases

We should be able to clean up the corresponding job. 

Maybe [setting the time to live (TTL) field of a Job resource](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) is a good solution.

kycheng:
@stonezdj any suggestions on this issue?

PR #1600, opened 2023-09-07
add ttlSecondsAfterFinished for migration-job

kycheng:
Signed-off-by: kychen <kychen@alauda.io>

#1599

### DIFF
--- a/templates/core/core-pre-upgrade-job.yaml
+++ b/templates/core/core-pre-upgrade-job.yaml
@@ -12,6 +12,7 @@ metadata:
     "helm.sh/hook": pre-upgrade
     "helm.sh/hook-weight": "-5"
 spec:
+  ttlSecondsAfterFinished: 300
   template:
     metadata:
       labels:


stonezdj (approved):

kycheng:
@zyyw Can you help me review this PR?

zyyw (approved):

zyyw (commented):
According to this [doc](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases), there are two options that may achieve the purpose of deleting the job after the hook is successfully executed:
* [add a custom helm.sh/hook-delete-policy annotation](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies)
* [set the time to live (TTL) field of a Job resource](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/)

Should we use the first option to only delete the job when the hook is completed successfully?  Therefore, users can still investigate the job log if anything goes wrong.

kycheng:
> maybe try the `annotation` option?

Yes, it can also be handled by annotation. The current default behavior is before-hook-creation. In the case of uninstallation, several strategies supported by helm do not seem to work well.

The problem I want to solve is that there should be no resources left after uninstalling.

Issue #1609, opened 2023-09-28
fix: varying checksum


jjangga0214:
checksum(annotation) is automatically changed every time even when it should not.

For example, `checksum/secret` changes when a value regardless of it changes.

I think it should be fixed.

tobiwanobi:
This is still not working for me when deploying the 1.14.0 helm chart with ArgoCD v2.9.3

Everytime I commit anything to my repo these values get updated and a new replicaset is created.

PR #1612, opened 2023-10-06
Fix random values regenerating

Kajot-dev:
- Fixes regeneration of random secrets
- Adds support for existingSecret for `swift` and `oss`

Fixes: #1609 #1549

Essentially does the same thing as #1550 but with different approach


templates/registry/registry-dpl.yaml
```diff
@@ -120,6 +120,33 @@ spec:
...
+              optional: true
+        {{- end }}
+        {{- if .Values.persistence.imageChartStorage.oss.existingSecret }}
+        - name: REGISTRY_STORAGE_OSS_ACCESSKEYSECRET
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.persistence.imageChartStorage.oss.existingSecret }}
+              key: REGISTRY_STORAGE_OSS_ACCESSKEYSECRET
+              optional: true
+        {{- end}}
```
zyyw, line 149: We may also need to add this code block to `registryctl` container within `registry.dpl` deployment.

Kajot-dev: @zyyw You're correct. Fixed!

Kajot-dev:
@zyyw Reverted generating internal database password as this could possibly break existing installations and changes current behavior which allows empty password for internal database.
Also resolved conflicts.

values.yaml
```diff
@@ -291,6 +291,8 @@ persistence:
       username: username
       password: password
       container: containername
+      # keys in existing secret must be REGISTRY_STORAGE_SWIFT_PASSWORD, REGISTRY_STORAGE_SWIFT_SECRETKEY, REGISTRY_STORAGE_SWIFT_ACCESSKEY
```
zyyw, line 294: Please add an entry in `README.md` for this item.

zyyw (approved):

ywk253100 (approved):

kevin7s-io:
@OrlinVasilev,

Seeing the same in ArgoCD. It's not clear to me why the checksum is changing as diffing the ``{release}-harbor-core`` secret shows no differences between sync runs.

FWIW, I ended up just ignoring the checksum annotations in the ArgoCD app (you can be more tactical than this -- I'm lazy):

```yaml
  ignoreDifferences:
    - group: "*"
      kind: Secret
      jsonPointers:
        - /data
    - group: apps
      kind: Deployment
      jsonPointers:
        - /spec/template/metadata/annotations
```

PR #1607, opened 2023-09-27
fix external redis existing secret url render

MinerYang:
fix: https://github.com/goharbor/harbor/issues/19387

PR #1614, opened 2023-10-11
fix external redis existing secret url render

MinerYang:
cherry-pick of : https://github.com/goharbor/harbor-helm/pull/1607
fix: https://github.com/goharbor/harbor/issues/19387

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -144,12 +144,26 @@ app: "{{ template "harbor.name" . }}"
   {{- end }}
 {{- end -}}
 
+
+{{- define "harbor.redis.pwdfromsecret" -}}
+  {{- (lookup "v1" "Secret"  .Release.Namespace (.Values.redis.external.existingSecret)).data.REDIS_PASSWORD  | b64dec }}
+{{- end -}}
+
+{{- define "harbor.redis.cred" -}}
+  {{- with .Values.redis }}
+    {{- if (and (eq .type "external" ) (.external.existingSecret)) }}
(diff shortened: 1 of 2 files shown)


ywk253100 (approved):

zyyw (approved):

Kajot-dev:
@MinerYang  This PR contains a bug, it will not use redis existingSecret for registry component (registry component uses "harbor.redis.password" which still does not use existingSecret

MinerYang:
> @MinerYang This PR contains a bug, it will not use redis existingSecret for registry component (registry component uses "harbor.redis.password" which still does not use existingSecret
registry will use password from registry env
https://github.com/goharbor/harbor-helm/blob/39befead99018ff6f1eb6068074eb094aa871f3f/templates/registry/registry-dpl.yaml#L224

Kajot-dev:
That's **registryctl** container, but not the **registry** container. I deployed version of the chart with this PR and caching in redis enabled and **registry** container was getting NOAUTH from redis, while **registryctl** was not

MinerYang:
> That's **registryctl** container, but not the **registry** container. I deployed version of the chart with this PR and caching in redis enabled and **registry** container was getting NOAUTH from redis, while **registryctl** was not

1. that env is for registry container, registryctl is 

> That's **registryctl** container, but not the **registry** container. I deployed version of the chart with this PR and caching in redis enabled and **registry** container was getting NOAUTH from redis, while **registryctl** was not

Thanks @Kajot-dev! Sorry for the late response and it  indeed another issue introduced by https://github.com/goharbor/harbor-helm/pull/1179

PR #1618, opened 2023-10-13
停止trivy更新DB，需要把环境变量TRIVY_SKIP_DB_UPDATE也置为true

AlawnWong:
停止trivy更新DB，需要把环境变量TRIVY_SKIP_DB_UPDATE也置为true

### DIFF
--- a/templates/trivy/trivy-sts.yaml
+++ b/templates/trivy/trivy-sts.yaml
@@ -93,6 +93,8 @@ spec:
               value: {{ .Values.trivy.ignoreUnfixed | default false | quote }}
             - name: "SCANNER_TRIVY_SKIP_UPDATE"
               value: {{ .Values.trivy.skipUpdate | default false | quote }}
+            - name: "SCANNER_TRIVY_SKIP_DB_UPDATE"
+              value: {{ .Values.trivy.skipUpdate | default false | quote }}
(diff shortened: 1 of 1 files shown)


templates/trivy/trivy-sts.yaml
```diff
@@ -93,6 +93,8 @@ spec:
               value: {{ .Values.trivy.ignoreUnfixed | default false | quote }}
             - name: "SCANNER_TRIVY_SKIP_UPDATE"
               value: {{ .Values.trivy.skipUpdate | default false | quote }}
+            - name: "SCANNER_TRIVY_SKIP_DB_UPDATE"
```
zyyw, line 96: Hi @AlawnWong, thanks for you contribute to harbor-helm.

I don't think we need to set `SCANNER_TRIVY_SKIP_DB_UPDATE` to `true` as well. Reason:
1. there is no env called `SCANNER_TRIVY_SKIP_DB_UPDATE` in https://github.com/aquasecurity/harbor-scanner-trivy
2. the existing `SCANNER_TRIVY_SKIP_UPDATE` is used to call `--skip-db-update`
* https://github.com/aquasecurity/harbor-scanner-trivy/blob/main/pkg/etc/config.go#L35
* https://github.com/aquasecurity/harbor-scanner-trivy/blob/main/pkg/trivy/wrapper.go#L136

PR #1623, opened 2023-10-23
feat(nginx): add an option to set static ip for clusterIP service

Adibov:
I needed to create a clusterIP service for nginx which has static IP and there was no option to do so, therefore I created this MR to solve this issue.


zyyw:
Thanks for contributing to harbor-helm! Please undo the change in this file:
* Chart.yaml

BTW, also I see there is a large diff on `README.md`. It seems that you format it a little bit with only one new entry for `expose.clusterIP.clusterIP`, right?

zyyw:
Why your env is `static IP` specifically?

Adibov:
> Thanks for contributing to harbor-helm! Please undo the change in this file:
> 
> * Chart.yaml
> 
> BTW, also I see there is a large diff on `README.md`. It seems that you format it a little bit with only one new entry for `expose.clusterIP.clusterIP`, right?

I roll-backed changes on `Chart.yaml` file. Also, you're right about the changes on `README.md` file. I only add documentation about `expose.clusterIP.clusterIP` field.

Adibov:
> Why your env is `static IP` specifically?

Actually, we run self-hosted GitLab runners in a Kubernetes cluster in which we mounted the docker socket of all jobs' executors to their corresponding nodes. Therefore, we needed to access the harbor nginx service from nodes, and to achieve this functionality, we had to create a static IP service for harbor and a host alias in all nodes.

values.yaml
```diff
@@ -54,6 +54,8 @@ expose:
   clusterIP:
     # The name of ClusterIP service
     name: harbor
+    # The ip address of the ClusterIP service (leave empty for acquiring dynamic ip)
+    clusterIP: ""
```
zyyw, line 58: could you please change `clusterIP.clusterIP` to `clusterIP.staticClusterIP`?
Thanks

Adibov: Done

templates/nginx/service.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
 {{- end }}
 spec:
   type: ClusterIP
+  {{- if .Values.expose.clusterIP.staticClusterIP }}
+  clusterIP: {{ .Values.expose.clusterIP.clusterIP }}
```
zyyw, line 17: line 17. it should also be `.Values.expose.clusterIP.staticClusterIP` I suppose.

zyyw:
@Adibov , please rebase main branch code and resolve the conflicts in `README.md` file.
Thank you!

templates/nginx/service.yaml
Adibov: You're totally right, Fixed.

Adibov:
> @Adibov , please rebase main branch code and resolve the conflicts in `README.md` file. Thank you!

Rebased to the main branch and resolved conflicts. Any more comment would be welcomed :)

zyyw (approved):

wy65701436 (approved):
lgtm

Issue #858, opened 2021-02-11
[metrics] Prometheus & Grafana Integrations


dioguerra:
Can we add a metrics.serviceMonitor , metrics.prometheusRule , metrics.dashboard ?

I still didnt test the master chart, but this seems a reasonable inclusing for the v2.2.0 update

ywk253100:
You can configure the metrics in harbor chart 1.6.0(OSS Harbor v2.2.0): https://github.com/goharbor/harbor-helm/blob/master/values.yaml#L735

dioguerra:
This is not the same thing that i said...

dioguerra:
I said, among others also, to include a CM for harbor visualization
Installed by CM

https://stackoverflow.com/questions/55830850/how-to-import-custom-dashboards-to-grafana-using-helm

reasonerjt:
IMO, that is out of the scope of this repo.

PR #1635, opened 2023-11-02
Add Prometheus default alerts and Grafana health dashboard

Allex1:
Addresses https://github.com/goharbor/harbor-helm/issues/858

### DIFF
--- a/templates/metrics/dashboard.yaml
+++ b/templates/metrics/dashboard.yaml
@@ -1,0 +1,2551 @@
+{{ if and .Values.metrics.enabled .Values.metrics.serviceMonitor.enabled .Values.metrics.grafanaDashboards.enabled }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    grafana_dashboard: "1"
+    app: {{ .Release.Name }}
+  name: harbor-dashboard
+  namespace: {{ .Release.Namespace }}
+data:
+  harbor.json: |-
+    {
+      "__inputs": [],
+      "annotations": {
+        "list": []
+      },
+      "description": "Harbor",
+      "editable": true,
+      "gnetId": null,
+      "hideControls": false,
+      "id": null,
+      "links": [],
+      "panels": [
+        {
+          "cacheTimeout": null,
+          "datasource": {},
+          "description": null,
+          "editable": true,
+          "error": false,
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "palette-classic"
+              },
+              "custom": {
+                "axisLabel": "",
+                "axisPlacement": "auto",
+                "barAlignment": 0,
+                "drawStyle": "line",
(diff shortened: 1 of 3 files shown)


Allex1:
@zyyw pls review ?

zyyw:
Hi @Allex1,

Thanks for contributing to harbor!
On the one hand, these two manifests introduced by this PR can be created manually and separately even if it's not merged in the harbor-helm project; on the other hand, this feature could enrich the harbor functionalities in terms of monitoring harbor application. Would you be able to continue supporting this feature if there are issues open regarding it?

Currently, we'll keep this PR open and collecting the demand/feedback of this feature from the community.
Thanks again!

Allex1:
> Hi @Allex1,
> 
> Thanks for contributing to harbor! On the one hand, these two manifests introduced by this PR can be created manually and separately even if it's not merged in the harbor-helm project; on the other hand, this feature could enrich the harbor functionalities in terms of monitoring harbor application. Would you be able to continue supporting this feature if there are issues open regarding it?
> 
> Currently, we'll keep this PR open and collecting the demand/feedback of this feature from the community. Thanks again!

@zyyw I agree these can be side-loaded (and that's what we currently do) but these are under feature flag so there are no braking changes. 
Also by having them as part of the harbor-helm project could provide tremendous benefits from the community updates/changes as opposed to manage them internally .
Feedback gathering is fine I guess but that's why I linked to an older issue that was requesting this feature.
Thanks!

Allex1:
@zyyw can we merge this ?

Vad1mo:
@Allex1 as @zyyw correctly mentioned, can you continue to maintain and expand this part in the helm chart?
If so, we can add this functionality to harbor, If not, we should add this to the Wiki as a starting point for users.

Allex1:
@zyyw @Vad1mo I think this should be merged as a starting point for users nevertheless but I can also maintain it.
We had to sub chart this internally and add these resources separately in order to get some monitoring enabled on our stack. Having this ootb should be a good step forward.
Thanks

MinerYang:
We are now not an all-in-one services, and it is introduced the third-party `Prometheus` CRD which we do not have dedicated  resources to maintain it for long term. So we would like to hold this feature.

Allex1:
@MinerYang This is under a feature flag and disabled by default so it's backwards compatible. This is a very common practice in multiple OSS helm charts. In case users want to enable this they will be required to deploy the `PrometheusRule` object themselves. I can add this comment in needed.

volker-raschek:
Hello everyone, as a longtime maintainer of kubernetes environments, I can only endorse X's pull request. 

I agree that this practice is common and would be very happy to see the changes again in a next release of the chart. This would not only reduce complexity for the environment I maintain, but also for anyone who would like to monitor their system without much effort.

Allex1:
closing in favor of https://github.com/goharbor/harbor-helm/pull/1703

PR #1659, opened 2023-12-08
define different deploy strategy for jobservice

gminiba:
We need to be able to define different deployment strategy for JobService as it has a PV and does not work well with RollingUpdate. In contrast, other Deployments need to be redeployed without stopping service.


gminiba:
@zyyw would you please review?

Vad1mo:
you need to also update the vales and README to make the change clear.

gminiba:
@Vad1mo I have made the changes requested

gminiba:
will do the new one

Issue #1660, opened 2023-12-12
Support for Security contexts in values.yaml


Maven35:
really need to be able to set security contexts so that i can enforce not running processes with UID 0, for harbor core it starts up with a UID 0 process. is there anything coming soon to be able to support this.

zyyw:
We have this PR:
* https://github.com/goharbor/harbor-helm/pull/1695

merged in to support security contexts in values.yaml

Issue #1253, opened 2022-08-08
deploy harbor on openshift


Meza1923:
Hey, 

I'm trying to deploy harbor on OpenShift and make him use the certificate of the cluster.
When I set the next values in order to achieve that, no route is being created.

expose.type=ingress
expose.tls.enabled=true
expose.tls.certSource=none

Could someone help with that problem?

Thanks a lot,
Guy.

zyyw:
We currently not test harbor deployment on openshift.

arjunprasad2143:
Could you share the details of the version of the openshift used and the version of harbor helm.
For me it works without an issue

Meza1923:
> Could you share the details of the version of the openshift used and the version of harbor helm.
> For me it works without an issue

Openshift 4.8.19
Helm chart 1.9.2

MallocArray:
I'm on Openshift 4.10 and using chart 1.9.1 and 1.9.3 without issue.

We do use a certSource of "secret" and have our own TLS secret created, as well as the expose.ingress.hosts.core set to a shorter URL, so we just have harbor.domain.com and a cert that is valid for that, but the routes get created as expected in that setup.

I haven't used a default tls cert on ingress, so not sure if that is specifically where your problem lies.

Kajot-dev:
@Meza1923  By "the certificate of the cluster" do you mean certificate served by default?

First if the Routes are not created, that means, that something with your Ingress is not quite right.
I think I see the problem. Setting `expose.tls.enabled=true` and `expose.tls.certSource=none` set's `secretName` in `Ingress` tls section to `""`.

In order to enable default tls in OpenShift you will need to:
- Set `expose.tls.enabled` to `false`
- Add `route.openshift.io/termination` annotation to your `Ingress` with the value `edge` if you are not using internalTLS in Harbor (set `expose.ingress.annotations`)
- If you ARE using internalTLS in Harbor, Add `route.openshift.io/termination` annotation to your `Ingress` with the value `reencrypt` and set `route.openshift.io/destination-ca-certificate-secret` annotation accordingly (although I believe this is available in newer version of Openshift)

Here is an *example* configuration:
```yaml
expose:
  type: ingress
  tls:
    enabled: false
  ingress:
    hosts:
      core: my.domain
    annotations:
      route.openshift.io/termination: "edge"
```

fvogl:
Would be great if you could adjust the helm chart to work on Openshift 4.13. 

[...]

PR #1666, opened 2023-12-22
Add option to configure securityContext on both pods and containers

Kajot-dev:
Fixes: #1253 
Fixes: #1660 

# Why is it needed?

With the introduction of **PodSecurityStandard** (see https://kubernetes.io/docs/concepts/security/pod-security-standards/) in Kubernetes, users are able to enforce stricter security policies. In addition to that security-hardened distributions of k8s (like Openshift/OKD) changes the default behavior in 4.12+ (due to the introduction of mentioned PSP and integration with existing SecurityContextConstraints) and requires more values in `securityContexts` to be defined.

Because of this, it's becoming **impossible for users with stricter security to deploy the chart without forking it.**

# What is does?

- Adds option to configure `securityContexts` through the chart
- Does not change the default behavior ~(except for permission initContainer in internal database deployment, see below)~
- Has a clear indication that changing the user id/fs group is **not supported by official images** and may break permissions
- ~Makes `data-permissions-ensurer` initContainer in internal database deploy always run as a root, since it should be able to use `chmod` (if uid via `podSecurityContext` was set to `999` before, was it even working at all?)~

**Note:** This PR does not make the chart compatible with stricter security k8s distribution by default, but enables users to do so on their own.

If the idea gets accepted, I will also update the README.


Vad1mo (approved):

Kajot-dev:
Ok, I've checked and root is not necessary for `chmod` since it will check your permissions and elevate to root by itself.

MinerYang:
Hi @Kajot-dev ,

Could you elaborate more we should both have pods level and container level security contexts? Since almost all of the harbor pods only have one containers. And for registry pod , although it includes `registry` and `registryctl` containers, but both of them share one PV and same permission settings for them.

Kajot-dev:
> Hi @Kajot-dev ,
> 
> Could you elaborate more why should we both have pods level and container level security contexts? Since almost all of the harbor pods only have one containers. And for registry pod , although it includes `registry` and `registryctl` containers, but both of them share one PV and same permission settings for them.

@MinerYang Od course! Ability to modify `securityContext` on the Pod scope is not necessarily required. Main goal of this PR is for user to be able to comply with PSP `restricted` (which requires additional fields in `securityContext` on the Container scope, https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted). However #1673 achieves the same goal without exposing configuration to the users and automatically forcing most safe/strict security.

In other words if #1673 gets merged, I am satisfied and this PR can be closed.

MinerYang:
Thanks @Kajot-dev 
Will close this PR and please help to review https://github.com/goharbor/harbor-helm/pull/1695 as we discussed.

Issue #456, opened 2019-12-02
Support configuring security context


ywk253100:
Currently, the containers in Harbor run as specific user(user ID 10000 or 999), and the user cannot be changed because some directives such as https://github.com/goharbor/harbor/blob/release-1.10.0/make/photon/registry/Dockerfile#L9. This prevents the supporting for configuring`runAsUser`. 
For other security context directives https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies, we need more test to verify they can work as expected.

There is already a PR https://github.com/goharbor/harbor-helm/pull/266 from community for the security context supporting

NickHes:
@ywk253100 Any plans to support this in the future? With Kubernetes 1.24 reaching EOL and clusters migrating to 1.25 with the new policies I suspect the need for this will grow. Currently we cannot run Harbor with the new policies enabled. If you are open for PR's for this we could most certainly take a look.

PR #1673, opened 2024-01-03
Add securityContext definitions on container level

rgarcia89:
Deployment hardening by adding / updating the securityContext definitions on container level.


Vad1mo:
what are the default settings, if this is not set?

The reason I am asking is for backwards compatibility

rgarcia89:
@Vad1mo default values are as followed on container level:

`privileged` is false per default
`allowPrivilegeEscalation` is false if privileged is set to false
`capabilties` defaults to the set of capabilities granted by the container runtime

Vad1mo (approved):

Kajot-dev:
Having that would **partially** solve the problem of non-compliance with restricted `PodSecurityStandard`, since it requires `seccompProfile` definition (see https://kubernetes.io/docs/concepts/security/pod-security-standards/#privileged) in 1.19+.
Could `seccompProfile` be added for k8s 1.19+ (using helm `Capabilities`)?

Kajot-dev:
@rgarcia89 `initContainers` also should have these and container from `core-pre-upgrade-job.yaml` should also have that.

If these changes would get merged (with `seccompProfile` and missing containers/initContainers) If `seccompProfile` I would favour this PR over my #1666, since all containers would comply with the most restrictive k8s security standard.

rgarcia89:
@Kajot-dev valid point. I will add the `seccompProfile` and also add the `securityContext` definition to the `initContainers` and the once I missed.

rgarcia89:
Perfect, so we only need to find a second reviewer.
Maybe @zyyw?

MinerYang:
https://github.com/goharbor/harbor-helm/issues/456
We are in review of this PR. Will merge it once it done

Kajot-dev:
@rgarcia89 Would it be possible to add `runAsNonRoot` to `secuurityContext` one the pod scope since
- It already runs as non-root
- PSP `restricted` requires this field to be present (https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted)

rgarcia89:
@Kajot-dev sure - I just added them.

MinerYang:
Hi @rgarcia89 , @Kajot-dev 

Thanks for both of your contributions on the PSP and security contexts. By reviewing both of pt-1763 and pr1666, I have a few thoughts combined with some of our downstream usage scenario:
- how about we still expose a `containerSecurityContext` like @Kajot-dev provided in his PR. But not set for every container respectively, instead using only one settings for all containers in values.yaml, e.g. `Values.containerSecurityContext`
- and for this containerSecurityContext settings, we have default values same in these two PRs to comply with PSP

How about we focus on this PR and I will close the other instead(https://github.com/goharbor/harbor-helm/pull/1666).

rgarcia89:
@MinerYang imho, I'm fine with both approaches. I'm fine with making them configurable through the values file, just as much as hardcoding them into the specific YAMLs. I opted for the latter initially, believing there isn't much reason for anyone to modify them. Nevertheless, if you prefer them to be modifiable, I propose we close my PR and proceed with yours. My main goal is to ensure that we integrate this into the Helm chart, regardless of the method ;)

MinerYang:
Thanks @rgarcia89 .
for now I think what we hard coding in the PR is enough for most of the cluster providers. But I still left it as configurable in case of Kubernetes evolve these standards in the future and also make it more flexible.

MinerYang:
Thanks @rgarcia89 !

Will close this PR and please help to review https://github.com/goharbor/harbor-helm/pull/1695 as we discussed.

PR #1675, opened 2024-01-03
fix: avoid decoding plain text password when database secret exists

### LINKED ISSUE #1674: Misbehaviour during upgrade
When using the internal database, during an upgrade from a previous Chart version, `POSTGRES_PASSWORD` gets replaced with `illegal base64 data at input byte 3` this is caused by a075633 and specifically by line 101:

`{{- .Values.database.internal.password | default (index $existingSecret.data "POSTGRES_PASSWORD") | b64dec -}}`

that causes an attempt to decode `.Values.database.internal.password` (which is in fact already plain text) instead of decoding only the existing secret value like this

`{{- .Values.database.internal.password | default (index $existingSecret.data "POSTGRES_PASSWORD" | b64dec) -}}`

barsa-net:
Fixes #1674


Vad1mo (approved):

mkjpryor:
Can we please get this merged and in a release ASAP? Currently the behaviour is incorrect on both upgrade and new installations.

ywk253100 (approved):

OrlinVasilev:
@barsa-net congrats on merging your first PR here :)))

PR #1687, opened 2024-01-19
Add condition to disable nginx even if not exposing via ingress

rgarcia89:


Vad1mo:
This is the same as https://github.com/goharbor/harbor-helm/pull/1073

we should unite those two PRs. 

@zyyw @ninjadq @wy65701436, we need such a PR more than ever more and more people are using different Ingresses nowadays, especially with GatewayAPI. Harbor can not support all the different options, so IMO we should add the option to not render the Ingress at all as we can not cover all the edge cases.

Vad1mo (changes requested):
we also need to update the README.

Vad1mo (approved):
awesome

README.md
```diff
@@ -78,6 +80,7 @@ The following table lists the configurable parameters of the Harbor chart and th
...
+| `expose.enabled`                                                            | Set to false if no `ingress`, `clusterIP`, `nodePort` or `loadBalancer` should be created or you plan to expose Harbor in a way not offered by this chart.                                                                                                                                                                                                                                                                                                                                  | `true`                                                                              |
```
wy65701436, line 83: Instead of adding a new item, I prefer to include additional Enums such as "None" for the expose type. This approach will not overly complicate the code.

rgarcia89: In my humble opinion, I believe that an enable setter aligns better with the current chart structure. 
internalTLS, persistence, trivy, metrics, and trace each have one. 
So, essentially, all dispensable components have one. Why deviate from this approach here now?

Vad1mo: I agree `enable = true/false` is common practice in Helm Chart development, we also use it in that chart. While it seem to be more verbose in code it is much easier and straightforward to understand for users what is going on and what is going to happen

MinerYang: Enabled more like a two sides thing, like we are using internalTLS or not, there's no third choice. But for `expose`, we have so many choices already, and made our control flow of this more redundant with both `type` and `enabled` . We could like to make our template simple and fulfill the requirements. 
I will take some time to maybe have an overall adjustment for this part, and please @rgarcia89 help  to review on it if your are willing to.

rgarcia89: I agree with @Vad1mo on that. I don't see why adding an enable switch should increase the complexity of the template. From my understanding, this is following a very simple approach and is not creating a misleading situation, which could be created by a `type: none` expose.

MinerYang: I agree it is not complex, and my first thoughts was could make it more simple when we only use `expose.type` as the entry-point instead of combine with `expose.type` and `nginx.enabled` would be more simple for users to understand the proxy usage.
maybe have something like this in `values.yaml`
```
expose:
    ## enum type Nginx, Ingress, ClusterIP,  NodePort, Loadbalancer, Others
    type:
    nginx:
        image:
        ...
    ingress:
       hosts:
       ...
    clusterIP:
        name:
        ...
    loadBalancer:
        name:
        ...
    nodeport:
        name:
        ...
```
While the templates would be  something like this
```
{{- if Values.expose.type == nginx }}
```

Sure I will take your comments into consideration, And please point me out if I have something understood wrong. Thx~

rgarcia89: Any update here?

Vad1mo:
@MinerYang @wy65701436, can we get this thing merged in? What is topping us from doing so?

rgarcia89:
Any update here?

Vad1mo:
@rgarcia89, can you resolve the conflicts, por favor?

rgarcia89:
@Vad1mo todo listo

PR #1693, opened 2024-02-05
Remove the key harbor section

wy65701436:
Remove the key “harbor” from the Ingress section and move the label to the upstream.


values.yaml
```diff
@@ -46,11 +46,8 @@ expose:
       ingress.kubernetes.io/proxy-body-size: "0"
       nginx.ingress.kubernetes.io/ssl-redirect: "true"
       nginx.ingress.kubernetes.io/proxy-body-size: "0"
-    harbor:
-      # harbor ingress-specific annotations
-      annotations: {}
-      # harbor ingress-specific labels
-      labels: {}
+    # ingress-specific labels
+    labels: {}
```
ywk253100, line 50: Please make sure the corresponding part in README is updated.

ywk253100 (approved):

zyyw (approved):

Vad1mo (commented):
This is a breaking change, we need to bump the major helm chart version. The changelog should address the migration that user need to do.

Kajot-dev:
@Vad1mo If you are going to do a major version bump, could we use this to consolidate persistence settings?

Currently persistence settings are as follows

- **registry** persistence is dependent on the persistence.enabled and imageChartStorage
- **trivy** persistence should be configured separetely since it can utilise `ReadWriteOnce` while running multiple replicas and other components can not
- **jobservice** persistence is dependent on jobLoggers
- **redis** persistence is not configurable

It's quite confusing to keep track of. Can't we just split it into components? Like, persistence would be configured separately for database, trivy, registry and jobservice. I think it would be less confusing than having one setting and make users track some internal chart logic whether PVC will be created or not.

If so, we can start a discussion in a separate issue and I can provide a pull request for that

PR #1539, opened 2023-07-11
Add support to run in the restricted psa policy

dwagelaar:
pods "harbor-core-..." is forbidden: violates PodSecurity "restricted:latest": runAsNonRoot != true (pod or container "core" must set securityContext.runAsNonRoot=true)

PR #1304, opened 2022-09-19
Add containerSecurityContext to components

funkypenguin:
Hey guys,

This PR attempts to address the issue raised here : https://github.com/goharbor/harbor-helm/issues/1286

Basically, it's desirable under some circumstances to allow the user to customize `.spec.container.securityContext`, to meet compliance requirements (_such as apply seccomp policy_). 

An [objection raised](https://github.com/goharbor/harbor-helm/issues/1286#issuecomment-1240344446) in #1286 was that allowing custom securityContext could cause breakage, since some of the apps' UIDs are hard-coded in their docker files. To this end, I've avoided changing the _pod_ securityContexts, and focused only on adding the ability to augment the _container_ securityContext.

This means that in the case of core, for example, the following hard-coded securityContext still applies:

```
      securityContext:
        runAsUser: 10000
        fsGroup: 10000
```

But a user could pass this in `values.yaml`:

```
core:
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault  
```

.. which would greatly improve our compliance / security posture, without impacting the intended effect of the hard-coded _pod_ securityContext.

PR #1695, opened 2024-02-06
Add security context

MinerYang:


Kajot-dev:
I have one comment. As far as I can see, database and redis use different uids and gids from the rest of the components, so I guess global `values.containerSecurityContext` would have to either:

-  **not** include `runAsUser` and `fsGroup` and these would be hardcoded in the manifests 
-  in case of database/redis use helm's `omit` like this `omit "runAsUser" "fsGroup"` and only have hardcoded uids/gids in in redis/database

MinerYang:
Hi @Kajot-dev ,

Thanks for your review and sorry for the late response since i just came back from holidays.
Good point! And indeed, none of our containers are running as root, however we need explicitly claim this `runAsNonRoot: true` to meet our restricted pod security standards.
What I initially want to write in values.yaml is a empty yaml looks like below and users could configure it on demand. Current version is just filled with a common template.
```
containerSecurityContext: {}
```

MinerYang:
relevant PRs:
- https://github.com/goharbor/harbor-helm/pull/1539
- https://github.com/goharbor/harbor-helm/pull/1673
- https://github.com/goharbor/harbor-helm/pull/1666
- https://github.com/goharbor/harbor-helm/pull/1304

Thanks @Kajot-dev , @rgarcia89 @dwagelaar , @funkypenguin for all of the contributions!

Let's focus on review this PR to meet several requirements.

templates/core/core-dpl.yaml
```diff
@@ -144,6 +144,9 @@ spec:
 {{- with .Values.core.extraEnvVars }}
 {{- toYaml . | nindent 10 }}
 {{- end }}
+        {{- if not (empty .Values.containerSecurityContext) }}
+        securityContext: {{ .Values.containerSecurityContext | toYaml | nindent 10 }}
+        {{- end }}
```
zyyw, line 149: @MinerYang How about the `core-pre-upgrader-job.yaml`, we may also need to apply the same change to that file as well.

values.yaml
```diff
@@ -1,3 +1,15 @@
...
+containerSecurityContext:
+  privileged: false
+  allowPrivilegeEscalation: false
+  seccompProfile:
+    type: RuntimeDefault
+  runAsNonRoot: true
+  capabilities:
+    drop:
+      - ALL
+
```
zyyw, line 12: move this new config to be here:
```
cache:
  # default is not enabled.
  enabled: false
  # default keep cache for one day.
  expireHours: 24

<<--- here

# If service exposed via "ingress", the Nginx will not be used
nginx:
```

wy65701436 (approved):
lgtm

templates/core/core-pre-upgrade-job.yaml
```diff
@@ -47,6 +47,9 @@ spec:
               secretKeyRef:
                 name: {{ .Values.database.external.existingSecret }}
                 key: password
+        {{- end }}
+        {{- if not (empty .Values.core.containerSecurityContext) }}
+        securityContext: {{ .Values.core.containerSecurityContext | toYaml | nindent 10 }}
```
Kajot-dev, line 52: Shouldnt this use `.Values.containerSecurityContext` instead of `.Values.core.containerSecurityContext`?
`.Values.core.containerSecurityContext` does not exists in `values.yaml`

Kajot-dev (changes requested):
Just one thing in core-pre-upgrade job, the rest looks good to me

templates/core/core-pre-upgrade-job.yaml
rgarcia89: good catch :+1:

zyyw (approved):

PR #1699, opened 2024-02-07
remove enableMigrateHelmHook

wy65701436:
Set enableMigrateHelmHook always be true to avoid any migration failure, especially in the case of large mount data.


Vad1mo (changes requested):
It is not clear why you think this is the better option, and should be the default. 

Could you kindly provide further details regarding the issue it addresses? Are there any references to open issues?

wy65701436:
The purpose behind introducing this setting was to address situations where migration takes more than 60 seconds to complete. When users encounter this issue, they are required to enable this option and initiate the migration process again.

There is no harm in enabling it by default, and there is no need to include this option in the values.yaml file for users to decide whether to enable it or not.

Vad1mo (approved):

PR #1702, opened 2024-02-08
Add support for providing an existing secret to internal database

cten:
This PR provides the means to use an external tool to create/manage secrets that then can be used by harbor internal database instead of relying on helm auto-generated password.

### DIFF
--- a/templates/database/database-ss.yaml
+++ b/templates/database/database-ss.yaml
@@ -41,26 +41,6 @@ spec:
       automountServiceAccountToken: {{ .Values.database.internal.automountServiceAccountToken | default false }}
       terminationGracePeriodSeconds: 120
       initContainers:
-      # as we change the data directory to a sub folder to support psp, the init container here
-      # is used to migrate the existing data. See https://github.com/goharbor/harbor-helm/issues/756
-      # for more detail.
-      # we may remove it after several releases
-      - name: "data-migrator"
-        image: {{ .Values.database.internal.image.repository }}:{{ .Values.database.internal.image.tag }}
-        imagePullPolicy: {{ .Values.imagePullPolicy }}
-        {{- if not (empty .Values.containerSecurityContext) }}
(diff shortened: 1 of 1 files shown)


Vad1mo (approved):
LGTM

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 1.4.0-dev
+version: 1.4.1-dev
```
zyyw, line 3: please do not change this.

templates/core/core-secret.yaml
```diff
@@ -21,7 +21,7 @@ data:
   {{- if not .Values.existingSecretAdminPassword }}
   HARBOR_ADMIN_PASSWORD: {{ .Values.harborAdminPassword | b64enc | quote }}
   {{- end }}
-  {{- if not .Values.database.external.existingSecret }}
+  {{- if and (not .Values.database.external.existingSecret) (not .Values.database.internal.existingSecret) }}
```
zyyw, line 24: need to update `core-dpl.yaml` as well

templates/exporter/exporter-dpl.yaml
```diff
@@ -82,6 +82,13 @@ spec:
               name: {{ .Values.database.external.existingSecret }}
               key: password
         {{- end }}
+        {{- if .Values.database.internal.existingSecret }}
```
zyyw, line 85: may need to change `{{- if .Values.database.internal.existingSecret }}` to `{{- if and (eq .Values.database.type "internal") .Values.database.internal.existingSecret -}}`

zyyw: also may also update line 78, `{{- if .Values.database.external.existingSecret }}` -> `{{- if and (eq .Values.database.type "external") .Values.database.external.existingSecret }}`

cten:
I had issues with merging upstream changes, so I recreated the PR here https://github.com/goharbor/harbor-helm/pull/1733

PR #1706, opened 2024-02-14
add multiple domains

nevlkv:
Harbor may use multiple domains for registry and helm charts

```
https://registry.<fqdn>
https://helm-charts.<fqdn>
```

and we needed patching ingress for operate in this state

Issue #1641, opened 2023-11-20
Helm Chart 1.13.1: Error with 'existingSecret' for External Redis Database


pierreblais:
## Issue Description

I have encountered a regression in the Helm Chart with the latest patch (1.13.1). Specifically, when attempting to provide an existingSecret for an external Redis database, the following error occurs:

~~~bash
$ helm template my-release harbor/harbor --values helm-values.yaml -n harbor --debug --version 1.13.1
install.go:194: [debug] Original chart version: "1.13.1"
install.go:211: [debug] CHART PATH: /home/vscode/.cache/helm/repository/harbor-1.13.1.tgz


Error: template: harbor/templates/trivy/trivy-sts.yaml:26:28: executing "harbor/templates/trivy/trivy-sts.yaml" at <include (print $.Template.BasePath "/trivy/trivy-secret.yaml") .>: error calling include: template: harbor/templates/trivy/trivy-secret.yaml:10:15: executing "harbor/templates/trivy/trivy-secret.yaml" at <include "harbor.redis.urlForTrivy" .>: error calling include: template: harbor/templates/_helpers.tpl:198:48: executing "harbor.redis.urlForTrivy" at <include "harbor.redis.url" $>: error calling include: template: harbor/templates/_helpers.tpl:166:64: executing "harbor.redis.url" at <include "harbor.redis.cred" $>: error calling include: template: harbor/templates/_helpers.tpl:155:25: executing "harbor.redis.cred" at <include "harbor.redis.pwdfromsecret" $>: error calling include: template: harbor/templates/_helpers.tpl:149:56: executing "harbor.redis.pwdfromsecret" at <.Values.redis.external.existingSecret>: nil pointer evaluating interface {}.REDIS_PASSWORD
[...]

PR #1712, opened 2024-02-22
add support for configuring internal domain

tomkukral:
This option can be beneficial if certificates aren't allowed to use bare domain.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -39,6 +39,12 @@ release: {{ .Release.Name }}
 app: "{{ template "harbor.name" . }}"
 {{- end -}}
 
+{{- define "harbor.internalDomain" -}}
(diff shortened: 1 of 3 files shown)


Vad1mo:
@tomkukral isn't  this going into the same direction as https://github.com/goharbor/harbor-helm/pull/1706?

tomkukral:
> @tomkukral isn't this going into the same direction as #1706?

Feature implemented in #1706 is different, IMHO.#1641 

* #1706 is adding support for multiple domain for external access (ingress). I like this change :)
* #1712 is enabling domain support for communication between components (internal TLS)

I'm running testing PoC on top of my branch so feel free to ping me if I should run any specific test.

zyyw:
Thanks for contributing to harbor-helm. This might be a too specific change. closing it now.

PR #1713, opened 2024-02-22
Add certmanager as internalTLS source

Kajot-dev:
Implements the same exact certificates for internalTLS as "auto" option, but using certmanager. 
This solved a couple problems:
1. Certificates are automatically renewed
2. Certificates are no longer regenerated each time on `helm upgrade`

Will test it on a working instance


whu-viasat (commented):
why aren't we merging this?

whu-viasat:
@Kajot-dev @MinerYang

PR #1728, opened 2024-03-11
Separate updateStrategy configuration for jobservice and registry deployments

jwojnarowicz:
Hello,

I am happy to share me Pull Request that addresses issues I encountered during working with Harbor Helm.

# What improvements does these changes make?

- Adds separation between update strategies for both persistent volume dependent deployments - `jobservice` and `registry`,
- When someone uses `s3` like backend, there is no need to change the strategy to `Recreate` as it does not depend on `PVC` and `RWM/RWO` access,
- When someone still has to use `Recreate` strategy for jobservice running on PVC in `RWO` and at the same time does not use `PVC` for registry storage, then `Recreate` creates unavailability of registry pods during update even though there is no such need.

I have also changed the README.md to reflect the changes. 
If you prefer I can also put these options inside `jobservice` and `registry` options then it would be `jobservice.updateStrategy.type` instead (and same for registry).

 Cheers 😄


jwojnarowicz:
@MinerYang can you check?

PR #1733, opened 2024-03-19
Support existing secret for internal DB

cten:


zyyw:
If it is an internal database, it should be managed by `database-secret.yaml`. I still don't get why you want to introduce existing secret for internal database.

PR #1751, opened 2024-04-24
Update readme.md: trivy default PVC size

dwrolvink:
changed outdated default.

Small change, but in our case it sent us for a loop because we looked here instead of just in the values.yml :)


dwrolvink:
Okay nevermind, I see I need to sign all my commits, if anyone reads this, please update the readme yourselves if you want.
Thanks a lot for this project! Cheers.

PR #1762, opened 2024-05-30
feat: manage the fsGroupChangePolicy configuration using the values.yaml 

wuzhihui1123:
Move the fsGroupChangePolicy value of the registry to the values.yaml, so that different field values can be configured for it.

### DIFF
--- a/templates/registry/registry-dpl.yaml
+++ b/templates/registry/registry-dpl.yaml
@@ -44,7 +44,9 @@ spec:
       securityContext:
         runAsUser: 10000
         fsGroup: 10000
-        fsGroupChangePolicy: OnRootMismatch
+        {{- if not (empty .Values.registry.fsGroupChangePolicy) }}
+        fsGroupChangePolicy: {{ .Values.registry.fsGroupChangePolicy }}
+        {{- end }}
 {{- if .Values.registry.serviceAccountName }}
       serviceAccountName: {{ .Values.registry.serviceAccountName }}
 {{- end -}}
--- a/values.yaml
+++ b/values.yaml
@@ -611,6 +611,7 @@ registry:
   nodeSelector: {}
   tolerations: []
   affinity: {}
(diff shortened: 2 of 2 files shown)


zyyw:
please raise PR against the `main` branch

zyyw:
https://github.com/goharbor/harbor-helm/blob/f547605c75badb6675810327cc4809b3ec4202a8/templates/registry/registry-dpl.yaml#L92

https://github.com/goharbor/harbor-helm/blob/f547605c75badb6675810327cc4809b3ec4202a8/templates/registry/registry-dpl.yaml#L231

users can configure it via the `containerSecurityContext`.

wuzhihui1123:
> https://github.com/goharbor/harbor-helm/blob/f547605c75badb6675810327cc4809b3ec4202a8/templates/registry/registry-dpl.yaml#L92
> 
> https://github.com/goharbor/harbor-helm/blob/f547605c75badb6675810327cc4809b3ec4202a8/templates/registry/registry-dpl.yaml#L231
> 
> users can configure it via the `containerSecurityContext`.

Certainly! Here is the translation:

---

Users can configure the spec.containers[*].securityContext vaule within a pod using `.Values.containerSecurityContext`, but the `spec.securityContext` under the pod cannot be configured. The `fsGroupChangePolicy` value under it is fixed to `OnRootMismatch`, which makes it impossible to deploy services in Kubernetes versions <=1.20. My modification allows the pod to support custom values for the `spec.securityContext` field, making it compatible with deployments in lower versions of Kubernetes.

Can you help reopen my pull request?

Vad1mo:
@wuzhihui1123 wir only Support k8s > 1.20

wuzhihui1123:
> @wuzhihui1123 wir only Support k8s >= 1.20. There is no intent to support older k8s version

  Well, it seems I have to fork my own repository.  😂   @Vad1mo

PR #1766, opened 2024-06-04
Make Ingress Optional

Vad1mo:
With the Upcoming [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/) and the wide adoption of different ingress controllers, eg. traefik, ALB etc. 

We should not aim to support all the options, but rather keep this chart flexible and lean. 

Hence I am proposing to make ingress optional with the flag `expose.enabled` `true/false`default is `true`


acelinkio:
I think this request is incredibly relevant.  I no longer use ingress controllers for exposing traffic to my cluster.   Any chance this could be reviewed?

janpipan:
You can disable the ingress object with current chart. If you set the `expose.type` to  `clusterIP`, the chart will deploy nginx server which will act as a API gateway for harbor services. You can then create HTTPRoute that has a `backendRefs` set to harbor service which is created when nginx server is deployed. If you would like I can provide you some examples of how we migrated our Harbor Ingress to Gateway API.

mts-dyt:
@janpipan Do we really need an internal NGINX instance if we want to expose Harbor through another method. This PR really makes sense, more than ever since the announce of NGINX retirement

PR #1768, opened 2024-06-05
add option to configure connMaxLifetime and connMaxIdleTime.

Vad1mo:
This adds the option to set Postgres connMaxLifetime and connMaxIdleTime on harbor core and with an override option on harbor exporter.


values.yaml
```diff
@@ -908,11 +908,18 @@ database:
     sslmode: "disable"
   # The maximum number of connections in the idle connection pool per pod (core+exporter).
   # If it <=0, no idle connections are retained.
-  maxIdleConns: 100
+  maxIdleConns: 50
```
zyyw, line 911: please don't change the default value of `maxIdleConns` and `maxOpenConns`. It has been working fine and it is consistent with the goharbor/harbor.

Vad1mo: updated the README, and setback the default to 100 and 900

zyyw:
@Vad1mo could you please help to resolve the conflicts? Thanks

zyyw (approved):

PR #1783, opened 2024-06-26
Support existing secret for external DB 

cest-pas-faux:
Hi,

This PR adds the ability to specify an existing secret for all the DB configurations (host, port, user, password, coreDatabase).

Tests done with this updated chart : 

- Reference an existing secret and setting `existingSecret.enabled: true`
- Hardcode the connection configuration values under `external` and set `existingSecret.enabled: false`

Both tests are successful, pull/push images OK.

### DIFF
--- a/templates/core/core-cm.yaml
+++ b/templates/core/core-cm.yaml
@@ -14,10 +14,12 @@ data:
     httpport = {{ ternary "8443" "8080" .Values.internalTLS.enabled }}
   PORT: "{{ ternary "8443" "8080" .Values.internalTLS.enabled }}"
   DATABASE_TYPE: "postgresql"
+  {{- if not .Values.database.external.existingSecret.enabled }}
   POSTGRESQL_HOST: "{{ template "harbor.database.host" . }}"
   POSTGRESQL_PORT: "{{ template "harbor.database.port" . }}"
   POSTGRESQL_USERNAME: "{{ template "harbor.database.username" . }}"
   POSTGRESQL_DATABASE: "{{ template "harbor.database.coreDatabase" . }}"
+  {{- end }}
   POSTGRESQL_SSLMODE: "{{ template "harbor.database.sslmode" . }}"
   POSTGRESQL_MAX_IDLE_CONNS: "{{ .Values.database.maxIdleConns }}"
   POSTGRESQL_MAX_OPEN_CONNS: "{{ .Values.database.maxOpenConns }}"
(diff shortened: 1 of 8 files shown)


MinerYang:
Hi @cest-pas-faux ,

I don't think these settings expect `password` need to be mounted as secret, configmap is enough for them.
And it will better align with our other values conventions

Best,
Miner

cest-pas-faux:
Hi @MinerYang, thanks for this review

The `password` stored as secret is the current behavior and has not been introduced by me, see templates below : 

- https://github.com/goharbor/harbor-helm/blob/main/templates/core/core-dpl.yaml#L130
- https://github.com/goharbor/harbor-helm/blob/main/templates/database/database-secret.yaml

I didn't want to create a breaking change by removing this secret which is shared by different components.
How to you want me to proceed ?

cest-pas-faux:
Hi @MinerYang , any updates on this ?

cest-pas-faux:
Hi @MinerYang , just a heads up :)

cest-pas-faux:
Hi @MinerYang , did you have any time to look into this PR ?

cest-pas-faux:
Hi @MinerYang , this feature is still needed.

cest-pas-faux:
Still needed

cest-pas-faux:
Still needed

JefeDavis:
@MinerYang I believe what @cest-pas-faux is trying to accomplish is support a use-case like cloudnative-pg, which generates a secret with all needed information to connect to the database upon creation, this is a pretty popular way of provisioning databases in kubernetes and should be supported by harbor.

@cest-pas-faux it may be easier to get this approved and merged if you remove the breaking change here. people may be using the existingSecret value which currently takes a string, you should be able to check to see if the value is a string or a map and revert to the old logic if it is a string

ildoc:
interested! I'm waiting to this feature to be merged to deploy harbor with external-secrets!

thank you @cest-pas-faux for the good work, let's hope @MinerYang will merge this pr as soon as possible

fredericrous:
thanks for the PR. imo this change is definitely needed to use the harbor helm chart in production

maxju:
We also need this feature.

cest-pas-faux:
Hi,
I honestly don't know why @MinerYang has been actively ignoring this feature for the past two years so I'm going to close it.
The code may be outdated since I didn't monitor the changes done to the chart on the main branch.
If anyone wants to open another PR, feel free to do so.

Issue #1801, opened 2024-07-26
`registry-dpl` template container's args only coincidentally work.


aznashwan:
The `registry-dpl`'s template currently only happens to coincidentally work with the upstream `goharbor/registry-photon` image because:

- the upstream image's Dockerfile [explicitly sets an `entrypoint.sh` script as the `ENTRYPOINT`](https://github.com/goharbor/harbor/blob/v2.11.0/make/photon/registry/Dockerfile#L18)
    - `entrypoint.sh` ignores all arguments and [always exec's `/usr/bin/registry_DO_NOT_USE_GC serve /etc/registry/config.yml`](https://github.com/goharbor/harbor/blob/v2.11.0/make/photon/registry/entrypoint.sh#L13)
- the `registry-dpl` template [only sets `args: ["serve", "/etc/registry/config.yml"]`](https://github.com/goharbor/harbor-helm/blob/main/templates/registry/registry-dpl.yaml#L98), which only coincidentally matches the exec call in `entrypoint.sh` and could cause issues if the upstream image's `entrypoint.sh` gets changed.

NOTE: The upstream image's [`entrypoint.sh` also runs `install_cert.sh`](https://github.com/goharbor/harbor/blob/v2.11.0/make/photon/registry/entrypoint.sh#L11), which will basically gets "lost" in the Helm chart setup if we directly set `command: /usr/bin/registry_DO_NOT_USE_GC` in this chart.

#### Possible fixes:

IMO the set of possible fixes, from "most to least correct" would be:
- making the `entrypoint.sh` from the upstream Dockerfile transparently pass all args it gets to `/usr/bin/registry_DO_NOT_USE_GC`, with it defaulting to `serve /etc/registry/config.yml` if none are given
- just setting `/home/harbor/entrypoint.sh` as the entrypoint in this Helm chart
- explicitly calling `install_cert.sh` from the Helm chart too (though that risks de-syncing with the upstream Dockerfile)

aznashwan:
> IMO the set of possible fixes, from "most to least correct" would be:

> making the entrypoint.sh from the upstream Dockerfile transparently pass all args it gets to /usr/bin/registry_DO_NOT_USE_GC, with it defaulting to serve /etc/registry/config.yml if none are given

I have taken the liberty to propose what I though was the best fix here: https://github.com/goharbor/harbor/pull/20778

MinerYang:
https://github.com/goharbor/harbor/pull/20778#issuecomment-2270847050

PR #1812, opened 2024-08-13
Remove ineffective 'args' from registry-dpl template.

aznashwan:
This patch removes the 'args' field from the 'registry-dpl' Deployment template, as is was completely ignored by the upstream registry-photon image's entrypoint.sh script which does not accept/process any arguments in any way.

Fixes: https://github.com/goharbor/harbor-helm/issues/1801


MinerYang (approved):
Lgtm

@aznashwan  Thanks for your contributions !

wy65701436 (approved):
lgtm

Issue #1620, opened 2023-10-16
GCE ingress controller doesn't support wildcards with pathType=Prefix


danielfm:
Not sure this is already being tracked somewhere else, but at least for me, the current version of this chart generates an invalid Ingress configuration.

If you use the `GCE` ingress controller, like this:

```yaml
expose:
  ingress:
    controller: gce
```

The Ingress template returns errors like this when applying:

```
Warning: path /api/* cannot be used with pathType Prefix
```

The solution here was to fork the chart and set `pathType: ImplementationSpecific`.

PR #1817, opened 2024-08-28
Fix issue #1620

torrefatto:
Hi,

I was hit by this, I see these errors in my ingress event logs
```
Warning  Translate  9m52s (x1702 over 7d7h)  loadbalancer-controller  Translation failed: invalid ingress spec: failed to validate prefix path /api/* due to invalid wildcard; failed to validate prefix path /service/* due to invalid wildcard; failed to validate prefix path /v2/* due to invalid wildcard; failed to validate prefix path /chartrepo/* due to invalid wildcard; failed to validate prefix path /c/* due to invalid wildcard; failed to validate prefix path /* due to invalid wildcard
```

I dug a bit and found #1620 and I figured these changes should fix the issue.

### DIFF
--- a/templates/ingress/ingress.yaml
+++ b/templates/ingress/ingress.yaml
@@ -2,6 +2,7 @@
 {{- $ingress := .Values.expose.ingress -}}
 {{- $tls := .Values.expose.tls -}}
 {{- if eq .Values.expose.ingress.controller "gce" }}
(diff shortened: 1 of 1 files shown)


Vad1mo:
What ingress are you using?

reasonerjt (approved):

Issue #1823, opened 2024-09-05
Add feature - Support existing secret to configure http_proxy and https_proxy


ClementJ35:
Hi,

Actually, all proxy settings are bond to values and deployed as env and envFrom configMaps.
As we want to hide username and password in our gitlab instance, we would like to use an existing secret that we could sync from external tools (as external secret operator).

I will make a PR soon, working here : https://github.com/ClementJ35/harbor-helm/tree/add-existing-secret-to-proxy

But first i would discuss how to implement it. I see two options :

1. Break the existent and pass the HTTP_PROXY in a secret, generated by helm or provided by user.
2. Use the existent env values and adapt it.

I'd say that as we have secret information, it should be a secret. But it will break possible third party tools used by persons to tweak pod environment variable and/or configmaps holding the proxy configuration.

Any income is welcome.

ClementJ35:
After revision, it seems that no_proxy doesn't have to be secret. So I'm leaving it out of secret scope, notably because of insertion of local services made by helm charts helper : https://github.com/goharbor/harbor-helm/blob/main/templates/_helpers.tpl#L295

MinerYang:
Could you give an example of your proxy url?

ClementJ35:
The proxy is setup as a secret like this : 

`apiVersion: v1
kind: Secret
stringData:
  HTTP_PROXY: http://username:password@proxy.example.com
  HTTPS_PROXY: https://username:password@proxy.example.com
`

ClementJ35:
Hi, as in the PR (#1825), I want to have this discussion. @MinerYang , did my previous message answered your proxy URL question ?

ClementJ35:
Givin'up. We worked this around.

PR #1824, opened 2024-09-05
Add feature : use existing secret to set HTTP_PROXY & HTTPS_PROXY

ClementJ35:
This PR aim to resolve issue #1823.
This is the non breaking implementation as describe in N°2 options in issues.
This is a work in progress.

PR #1825, opened 2024-09-06
Add feature : use existing secret to set HTTP_PROXY & HTTPS_PROXY

ClementJ35:
Replace #1824 as I messed up with git history with signoff
Aim to resolve issue #1823, the 2nd way exposed in the issue.


ClementJ35:
This is not stale, at least to me. Could you review this and tell me which options you could choose ?
This is really needed by our organization to use Harbor in production environment.

ClementJ35:
Givin' up, we worked this around.

PR #1826, opened 2024-09-17
Add oidc configuration

71g3pf4c3:
Added oidc configuration in helm chart. User can provide existing client secret or insert it explicitly in values. 
All configuration specified will be merged with configureUserSettings json and passed to CONFIG_OVERWRITE_JSON variable. 
Additionally json merge provides validation of configureUserSettings, preventing passing invalid json into it.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -579,3 +579,35 @@ app: "{{ template "harbor.name" . }}"
 {{- define "harbor.ingress.kubeVersion" -}}
   {{- default .Capabilities.KubeVersion.Version .Values.expose.ingress.kubeVersionOverride -}}
 {{- end -}}
+
+{{/* oidc configuration */}}
+{{- define "harbor.oidcConfigSecret" -}}
+	{{- if .Values.core.oidc }}
+{
+  "auth_mode": "oidc_auth",
+  "oidc_name": "{{ .Values.core.oidc.name }}",
+  "oidc_endpoint": "{{ .Values.core.oidc.endpoint }}",
+	"oidc_extra_redirect_parms": "{{ .Values.core.oidc.extraRedirectParms | default "{}" }}",
+  "oidc_client_id":  "{{ .Values.core.oidc.clientId }}",
+  "oidc_client_secret": "{{ .Values.core.oidc.clientSecret | default (include "harbor.oidc.rawClientSecret" .) }}",
+  "oidc_groups_claim":  "{{ .Values.core.oidc.groupsClaim | default "" }}",
+  "oidc_admin_group":  "{{ .Values.core.oidc.adminGroup | default "" }}",
+  "oidc_scope": "{{ .Values.core.oidc.scope }}",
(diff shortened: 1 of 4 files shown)


reasonerjt:
It was a decision not to expose authentication settings in values.yaml.

Similarly, we chose not to enable user to setup LDAP in values.yaml, such configuration can only be made by providing a JSON string in "configureUserSettings"

71g3pf4c3:
> It was a decision not to expose authentication settings in values.yaml, b/c this is considered "day-2" settings, that admin will handle it after Harbor is installed.
> 
> Similarly, we chose not to enable user to setup LDAP in values.yaml, such configuration can only be made by providing a JSON string in "configureUserSettings"

This feature can be helpful with fully automated deployments and gitops environments like flux. I also added functionality to use exitsting incluster secret values. 
configureUserSettings provides very limited range of configuration with just plain string field and without json validation or any validation at all, so i think my solution can be more convenient way to configure.

71g3pf4c3:
@reasonerjt please consider this feature, it helps in automated deployments regularly

Vad1mo:
> It was a decision not to expose authentication settings in values.yaml, b/c this is considered "day-2" settings, that admin will handle it after Harbor is installed.
> 
> Similarly, we chose not to enable user to setup LDAP in values.yaml, such configuration can only be made by providing a JSON string in "configureUserSettings"

Not so long ago, CONFIG_OVERWRITE_JSON was introduced, to support upfront Harbor configuration. 
It is not clear to me or to Harbor users why CONFIG_OVERWRITE_JSON should be a day-2 operation? 
Users are not using CONFIG_OVERWRITE_JSON for day-2, only for day-1 or to be precise for the initial setup. 

IMO it is also difficult to justify why 80% of harbor users, who install harbor via Helm Chats should not be able to use this feature,

lknite:
OIDC is not daytwo in the world of gitops ... the only reason one would think that is if they are not using flux/argocd/kapp ...

Vad1mo (approved):

templates/_helpers.tpl
```diff
@@ -579,3 +579,35 @@ app: "{{ template "harbor.name" . }}"
 {{- define "harbor.ingress.kubeVersion" -}}
   {{- default .Capabilities.KubeVersion.Version .Values.expose.ingress.kubeVersionOverride -}}
 {{- end -}}
+
```
Vad1mo, line 582: I can see the desire for a more specific and individualized configuration option, I also fear that this would that this would add more complexity to the chart and be less flexible. 

Our current solution with `CONFIG_OVERWRITE_JSON ` isn't the best design with embedded json makes it not better.  I would suggest to support a simple yaml to json converter  option,

71g3pf4c3:
@Vad1mo fixed unit test for trimmed json. Helm trims all the whitespace and line breaks, so it breaks tests

71g3pf4c3:
Created new https://github.com/goharbor/harbor-helm/pull/2160

PR #1843, opened 2024-10-17
Remove chartrepo location for nginx configmap

reasonerjt:
The chunk should be removed b/c it used to serve chart museum and it has been removed from Harbor


MinerYang (approved):
lgtm

wy65701436 (approved):
lgtm

PR #1849, opened 2024-10-27
feature: added namespace helper

Bugs5382:
So I am working on a project that puts Harbor inside the same namespace as the project, instead of being in default. I was looking trhough the helm chart and noticed that there was no way to override the namespace. Thus, I wrote this up. I tested it locally, and everything seems to be working A-OK!

Thanks!!!

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -31,11 +31,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 Preserve the default behavior of the Release namespace if no override is provided
 */}}
 {{- define "harbor.namespace" -}}
-{{- if .Values.namespaceOverride -}}
-{{- .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
-{{- else -}}
-{{- .Release.Namespace -}}
-{{- end -}}
+{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/* Helm required labels: legacy */}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -25,6 +25,19 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 {{- end }}
 
+{{/*
+Construct the namespace for all namespaced resources
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+Preserve the default behavior of the Release namespace if no override is provided
+*/}}
+{{- define "harbor.namespace" -}}
+{{- if .Values.namespaceOverride -}}
+{{- .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- .Release.Namespace -}}
+{{- end -}}
+{{- end -}}
+
 {{/* Helm required labels: legacy */}}
 {{- define "harbor.legacy.labels" -}}
 heritage: {{ .Release.Service }}
--- a/templates/core/core-cm.yaml
+++ b/templates/core/core-cm.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "harbor.core" . }}
-  namespace: {{ .Release.Namespace | quote }}
+  namespace: {{ template "harbor.namespace" . }}
   labels:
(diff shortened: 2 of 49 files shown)


Vad1mo:
@Bugs5382 the your PR does not have a [valid DCO](https://github.com/goharbor/harbor-helm/pull/1849/checks?check_run_id=32115429254)

Bugs5382:
@Vad1mo  Completed. Sorry about that.

templates/_helpers.tpl
```diff
@@ -25,6 +25,19 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 {{- end }}
 
+{{/*
+Construct the namespace for all namespaced resources
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+Preserve the default behavior of the Release namespace if no override is provided
+*/}}
+{{- define "harbor.namespace" -}}
```
Vad1mo, line 33: this is much easier:

{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}

Vad1mo (approved):

MinerYang:
Hi @Bugs5382 ,

Would  explicitly specify the namespace while you installing satisfy your requirements? like...
```
helm install harbor ... -n test
```

reasonerjt (changes requested):
Per latest comment by @MinerYang  
This PR probably should not be merged into upstream.
I'm blocking this PR until we have better clarity.

Bugs5382:
> Per latest comment by @MinerYang This PR probably should not be merged into upstream. I'm blocking this PR until we have better clarity.

So the reason why I request this change as if your doing a helm template output, even specifying the namespace, the default one gets applied. Which is kinda hard when I am using this chart with krustomize and everything else. I updated the chart.

Bugs5382 (commented):
Compelted.

templates/_helpers.tpl
Bugs5382: Compelted.

MinerYang:
Hi @Bugs5382 ,
I understood this is a limitation from helm. Thanks for your contribution and I propose a more organized way that align with our former template convention by using the "helper func". Please help to review, appreacite!
https://github.com/goharbor/harbor-helm/blob/6c6c68baf32147edfeb4f5657a7ac241ffc89279/templates/_helpers.tpl#L10

MinerYang:
Hi @Bugs5382 ,

Another thing may need your help and we would appreciate if you could add some UT test cases including this changes here:
https://github.com/goharbor/harbor-helm/tree/main/test/unittest.

Bugs5382:
@MinerYang I'll check it out today!

Vad1mo:
> So the reason why I request this change as if your doing a helm template output, even specifying the namespace, the default one gets applied. Which is kinda hard when I am using this chart with krustomize and everything else. I updated the chart.

we are too running `helm template .` and I don't see that behavior instead I see correct namespaces.

Kajot-dev:
> > So the reason why I request this change as if your doing a helm template output, even specifying the namespace, the default one gets applied. Which is kinda hard when I am using this chart with krustomize and everything else. I updated the chart.
> 
> we are too running `helm template .` and I don't see that behavior, instead I see correct namespaces.

Running `helm template .` with the `-n`/`--namespace` option works for me too

vpechor:
> What is the purpose of overriding a namespace? My question is: Is there a situation where you want to have a different release namespace than an actual install namespace?

In our environment, we use strict GitOps approach using ArgoCD. Most of installed applications are templated using helm template command which does not offer the possibility to use .Release variables as helm install does. Using -n/--namespace is an option for small projects, but is not viable for bigger helm charts, where each module gets its own namespace (e.g. when using umbrella helm chart). 
Additionally using ArgoCD to use helm install is an option, but not preferred for bigger projects as there's no direct control of what gets to production and has introduced us to some nasty bugs in production in the past (which is of course user issue).

namespaceOverride would be greatly appreciated.

Kajot-dev:
@vpechor It indeed will be problematic with umbrella helm chart. In the meantime I would suggest using `ApplicationSet` in ArgoCD instead of umbrella helm chart. Moreover this is a helm design that (mostly) each helm release corresponds to a single namespace and an option to set different namespaces for helm sub-charts is not provided explicitly.

Issue #1847, opened 2024-10-22
helm install should check the length of the specified csrf key 


wy65701436:
In the Harbor core, if the specified CSRF key does not meet the requirements, Harbor will generate a random string with a length of 32 characters. Consequently, if you scale the pods to more than one, requests will fail due to a CSRF key mismatch.

We should aim to address this issue early in the installation phase.

https://github.com/goharbor/harbor/issues/21060

PR #1853, opened 2024-11-04
add XSRFkey valication of 32 chars

MinerYang:
fix: https://github.com/goharbor/harbor-helm/issues/1847


MinerYang:
By leverage the benefit of the error handling , we decided to handle the error from the harbor-core runtime since helm has limitation to throw the validation error prior the installation , and it is more reasonable to having this error back to client instead of randomly generate ones for customer when the CSRF key mistakenly specified without 32 chars.

Issue #986, opened 2021-06-24
Harbor core and jobservice health checks timeouts


jowko:
We are using Harbor Helm Chart 1.6.2 (which contains Harbor v2.2.2) on Kubernetes 1.19.
Core and jobservice pods are restarting from time to time because of the timeouts on readiness and liveness probes:
![core](https://user-images.githubusercontent.com/20277750/123268472-581cdb80-d4fe-11eb-84ca-3580ba4e437f.PNG)
![jobservice](https://user-images.githubusercontent.com/20277750/123268478-58b57200-d4fe-11eb-8da7-b950b671b07a.PNG)
By default, timeout for health check if 1 second. When I run command below inside containers, sometimes it took a few seconds to respond (in most cases it respond quickly). 
`curl localhost:8080/api/v2.0/ping`
I don't know what causes random long response time for health checks. But because of the Harbor probe settings (failureThreshold of 2 for core and default 3 for job service), these pods are restarted frequently.

In my opinion, Harbor should either configure bigger timeout for these services our expose configuration via values.yaml for these probes.
In helm chart it can be done easily, when in values.yaml we place such section:
```
core:
  readinessProbe:
    failureThreshold: 2
    periodSeconds: 10
```
And then in our deployment definition we can put (example copied from my Helm chart, didn't tested for yml validity):
```
        readinessProbe:
          httpGet:
            path: /api/v2.0/ping
            scheme: {{ include "harbor.component.scheme" . | upper }}
            port: {{ template "harbor.core.containerPort" . }}
          {{- .Values.core.readinessProbe | toYaml | nindent 12 }}
```
Then any chart user can specify any options he/she wants.

ninjadq:
Hi, in most of env the local network timeout will not exceed 1 second, I don't think we shouldn't edit this config item.

jowko:
Most of requests in our environment also does not exceed 1 second, but from time to time it does. And when it happens 2 times in a row, then our service is restarted and this can occur multiple times per day. We don't need to change default timeout. We could make it configurable instead as in example above.

jowko:
Issue still exists

jowko:
Issue still exists

capabilitysupportservices:
[...]

PR #1855, opened 2024-11-06
feat: core service ready/liveness probes to be configurable

dylanhitt:
Recently ran into an issue on our cluster that has a lot of churn (we use karpenter and leverage spot instances pretty heavily). For some reason there are points when a new node in the cluster comes up and the AWS VPC CNI resyncs we see a bit of latency. I couldn't quite figure out why.  The 1 second timeout on the readiness probe procs fails quite a bit causing noise in our events monitoring as well as sometimes restarting the container when it seems otherwise healthy. 

The only way I found to solve the issue is to increase the `timeoutSeconds`. 

Let me know what you all think.

Cheers

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -317,6 +317,20 @@ app: "{{ template "harbor.name" . }}"
   {{- end -}}
 {{- end -}}
 
+{{- define "harbor.core.probe.httpGet" -}}
+httpGet:
+  path: /api/v2.0/ping
+  port: {{ template "harbor.core.containerPort" . }}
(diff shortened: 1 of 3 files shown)


MinerYang:
Hi @dylanhitt ,

Thanks for your contribution!
We understood you may have some latency from your Kubernetes infra level. However it is not a common level, could you try to maintain this config first at your side, and we will see if more users from community will needs this.

Best,
Miner

realrill:
Hi all. I think this me could solve the long ongoing #986 issue.
It feels like quite a lot of people would benefit from it, myself included.

Any chance that someone can review and merge this please?

PR #1876, opened 2024-11-26
rollingUpdate only allowed if type=RollingUpdate

gkramer:
https://v1-28.docs.kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/#DeploymentSpec


MinerYang:
Hi @gkramer ,

This [rollingUpdate line](https://github.com/goharbor/harbor-helm/blob/cccbc3f3c287e06c62271245dc7ae03c8f46753b/templates/jobservice/jobservice-dpl.yaml#L16) is for Rolling update config params is update strategy is not Recreate. And normally this strategy is either Recreate or Rolling Update.  So  I don't see any problem of this, and if we delete this line as you presents in the PR, the config for rolling update would lost as well.

MinerYang:
Hi @gkramer ,

Could you please add the sign-off?

PR #1890, opened 2024-12-13
Support Trivy's ignorePolicy

toanhminh0412:
This is a starter PR to support Trivy's `--ignore-policy` option as documented [here](https://trivy.dev/v0.22.0/vulnerability/examples/filter/)

Trivy currently isn't packaged with any `.rego` file, and there is no way to specify a list of policies in `.rego` format for Trivy to ignore when reporting vulnerabilities.

This PR provides an option in `values.yaml` to set a certain set of policies for Trivy to ignore. The option is `trivy.ignorePolicy`, which can be set to either `none`, `basic` or `advanced`. `basic` and `advanced` options are from [Trivy's example policies](https://github.com/aquasecurity/trivy/tree/v0.22.0/contrib/example_policy).

This PR provides hard-coded policies for ease of use, which does come with limitations. We might want to allow users to create custom policies. If there is a need for this, we can further develop this PR to support it.

### DIFF
--- a/templates/trivy/trivy-advanced-policies-cm.yaml
+++ b/templates/trivy/trivy-advanced-policies-cm.yaml
@@ -1,0 +1,105 @@
+{{- if .Values.trivy.enabled }}
+{{- if eq .Values.trivy.ignorePolicy "advanced" }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trivy-advanced-policies
+  namespace: {{ .Release.Namespace | quote }}
+data:
+  advanced.rego: |
+    package trivy
+
+    import data.lib.trivy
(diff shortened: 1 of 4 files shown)


toanhminh0412:
Hi @reasonerjt ,

Could you please take a look at this PR when you have a chance? Let me know if there's anything I can clarify or improve. Thanks in advance! 🙏

We need this feature asap for our Harbor instance and it would be great if this can be included into your repo.

dleske:
> This PR is being marked stale due to a period of inactivty. If this PR is still relevant, please comment or remove the stale label. Otherwise, this PR will close in 30 days.

Still interested in this and think it has value.

Issue #1671, opened 2024-01-02
Please use `helm` >= 3.10.0 when use harbor-helm 1.14.0


zyyw:
Please use `helm` >= 3.10.0 when deploying harbor-helm 1.14.0. Otherwise, it may occur an error similar to the one below:
```
Error: INSTALLATION FAILED: template: harbor/templates/registry/registry-secret.yaml:11:62: executing "harbor/templates/registry/registry-secret.yaml" at <include "harbor.secretKeyHelper" (dict "key" "REGISTRY_HTTP_SECRET" "data" $existingSecret.data)>: error calling include: template: harbor/templates/_helpers.tpl:44:41: executing "harbor.secretKeyHelper" at <.data>: wrong type for value; expected map[string]interface {}; got interface {}
```

PR #1897, opened 2024-12-24
fix: improve helm lookup fallback in case of helm template rendering

khanhngobackend:
This improve the case of using `helm template .` to statically render the resources. Without this, it might produce error similar to the following even with default `values.yaml`

```
Error: template: harbor/templates/registry/registry-secret.yaml:12:62: executing "harbor/templates/registry/registry-secret.yaml" at <include "harbor.secretKeyHelper" (dict "key" "REGISTRY_HTTP_SECRET" "data" $existingSecret.data)>: error calling include: template: harbor/templates/_helpers.tpl:59:41: executing "harbor.secretKeyHelper" at <.data>: wrong type for value; expected map[string]interface {}; got interface {}
```

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -56,8 +56,8 @@ app: "{{ template "harbor.name" . }}"
 
 {{/* Helper for printing values from existing secrets*/}}
 {{- define "harbor.secretKeyHelper" -}}
-  {{- if and (not (empty .data)) (hasKey .data .key) }}
-    {{- index .data .key | b64dec -}}
+  {{- if and (not (empty (.data | default dict ))) (hasKey (.data | default dict ) .key) }}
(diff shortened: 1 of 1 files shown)


Kajot-dev:
Nice, what helm version are you using. It seems the same as: https://github.com/goharbor/harbor-helm/issues/1671
Did you test with version 3.10 or higher?

khanhngobackend:
Hi @Kajot-dev , I tested with latest Helm v3.16.4

Kajot-dev:
Ok, then I think this PR is a good thing. Mabye I didn't make this resilient enough to configurarion changes my bad (I made the secretKeyHelper, so my bad 😅)

Kajot-dev (approved):

PR #2138, opened 2025-02-15
Template External Url

neevnuv:
## Description
This PR introduces templating support for the externalURL value inside the core-cm ConfigMap by wrapping it with tpl, allowing for dynamic evaluation of Helm templates.

## Changes
Updated core-cm to use {{ tpl .Values.externalURL . }} instead of a static value.

## Rationale
Enables users to reference other values or use Helm functions within externalURL.
Increases flexibility for dynamic configurations in different environments.


neevnuv:
Hi @MinerYang, I would love if you could give this PR a view 😁.

MinerYang:
Hi @neevnuv ,

Thanks for contribution to harbor-helm.

By getting some contexts of using `tpl` functions for helm template, the only "potential benefit" I could thought would be dynamically rendering the templating at the the runtime. However when you doing this, other configs like certificate for nginx tls config would be inconsistent and causing issues. And using the chain templates function would increasing our helm template complexity, from my perspective, a static rendering is enough for this value.

However, I will keep this PR open and let's see if we will have same requirements from the community in the future. In the meantime, you could use this at your own fork repository.

Best,
Miner

neevnuv:
Hi @MinerYang,

Thanks for your feedback! I understand the concerns about consistency and complexity.

We use ArgoCD on OpenShift to deploy Harbor across multiple clusters via ApplicationSets. Each cluster’s externalURL must follow harbor.apps.<clusterName>.example.com, and we use .Values.global.clusterName for templating. This PR enables us to set externalURL dynamically, reducing manual effort and misconfigurations.

Notably, the only appearance of externalURL is in the core-cm.yaml ConfigMap, minimizing the risk of inconsistencies with other configurations like TLS.

While tpl adds some complexity, it simplifies multi-cluster management. Would you consider making this feature optional? It could benefit other users with similar needs. Looking forward to your thoughts!

MinerYang:
> Hi @MinerYang,
> Notably, the only appearance of externalURL is in the core-cm.yaml ConfigMap, minimizing the risk of inconsistencies with other configurations like TLS.
> 

Thanks @neevnuv for more details here. I will try to catch up more about ArgoCD and multi-cluster things about templating config.
A mainly concern here is once we dynamically change the `externalURL` here , we also need dynamically rotate the tls cert as well , so would you like to share how could we handle this properly?

PR #2144, opened 2025-02-24
harbor-helm: trivy should use secret generated by helm only if intern…

qdrddr:
When enabled trivvy with external redis, the harbor helm-chart incorrectly handles secrets and the manifest inflation stage.

Example of the helm-value file when the issue manifests, with helm chart version 1.16.2:
```yaml
trivy:
  enabled: true
redis:
  type: external
  external:
    addr: "myredis-svc:6379"
    existingSecret: "harbor-redis-mysecret"
```

The issue seems to be in the `harbor/templates/_helpers.tpl` with the definitions of `harbor.redis.cred` &`harbor.redis.pwdfromsecret` which tries to access Redis secret regardless if redis is deployed as external or internal. While it works fine when Redis is deployed internally, it fails when Redis is external since the secret does not exist in the inflation manifests by helm-chart, but will be accessible by the app when it is deployed in k8s.

trivy should use a secret generated by helm only if internal Redis is used in the value file, if external Redis is used, the chart should not try to pull the secret from the manifest.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -173,13 +173,19 @@ app: "{{ template "harbor.name" . }}"
 
 
 {{- define "harbor.redis.pwdfromsecret" -}}
-  {{- (lookup "v1" "Secret"  .Release.Namespace (.Values.redis.external.existingSecret)).data.REDIS_PASSWORD  | b64dec }}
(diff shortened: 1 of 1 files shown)


qdrddr:
https://github.com/goharbor/harbor-helm/issues/2148

PR #2153, opened 2025-03-18
Fix Redis URL formatting in Harbor Helm chart

Eyalm321:
Fixed Redis url templating issues for external redis kubernetes deployment. It wont recognize redis.cache.svc.cluster.local as a number anymore.
I havent done implementation for redis.external.host and redis.external.port since Ive seen it only on online goharbor helm chart and not actual implementation. let me know if you want that in there.


Eyalm321:
It might need some adjustment for tls. I couldnt get it to work eventually with redis

PR #2160, opened 2025-03-27
Feat/add.OIDC configuration

71g3pf4c3:
Added oidc configuration in helm chart. User can provide existing client secret or insert it explicitly in values.
All configuration specified will be merged with configureUserSettings json and passed to CONFIG_OVERWRITE_JSON variable.
Additionally json merge provides validation of configureUserSettings, preventing passing invalid json into it.

Previous - https://github.com/goharbor/harbor-helm/pull/1826
was stalled due to no activity from maintainers, please consider merging it again

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -579,3 +579,35 @@ app: "{{ template "harbor.name" . }}"
 {{- define "harbor.ingress.kubeVersion" -}}
   {{- default .Capabilities.KubeVersion.Version .Values.expose.ingress.kubeVersionOverride -}}
 {{- end -}}
+
+{{/* oidc configuration */}}
+{{- define "harbor.oidcConfigSecret" -}}
+	{{- if .Values.core.oidc }}
+{
+  "auth_mode": "oidc_auth",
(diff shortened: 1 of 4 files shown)


71g3pf4c3:
A bit later i will add fully fledged CONFIG_OVERWRITE_JSON generation or add template for downstream charts

MinerYang:
Hi @71g3pf4c3 ,

Thanks for connecting with us, by getting more context of this requirement, I have a few doubts that  we 've been discussed on the last PR.
 - for json validation, would it sufficient by adding `fromJson` directly to the current config like `CONFIG_OVERWRITE_JSON: {{ .Values.core.configureUserSettings | fromJson | ... }}`  ?
 - I understood there's gitops requirements for setting the OIDC configurations while installing the harbor instance,  But if we are having requirements like this for `LDAP`, or 'audit log', and we set every of these type of config a separate block of overwrite, it would increasing the complexity of merge and templating.
 
 
 So  I would like to get more info from your side, what is your pain point of using current `configureUserSettings` explicitly? Would it be better to have an overall `existingSecretOfconfigureUserSettings` if you don't want to use a plain text in the values.yaml? 

Best.
Miner

PR #2161, opened 2025-03-27
Add Volume Mounts, ServiceAccount, and Custom Resource Support

urpylka:
This PR enhances the Harbor Helm chart by:

- Enabling volume declaration and mounting in the core and registry components (e.g., for a secret provider class).
- Introducing a ServiceAccount resource to complement the existing serviceAccountName parameter.
- Allowing users to define arbitrary custom resources via values.yaml (inspired by the ArgoCD approach), such as SecretProviderClass definitions.

These improvements increase the chart's flexibility and align it with modern deployment practices.

---

In addition, I noticed that many lines contain arbitrary trailing spaces. I fixed these issues where I encountered them, but it appears that you should enforce a mandatory code linter in our CI process for all PRs. This will help ensure that the code remains clean and consistent, making it easier and more pleasant for everyone to review.


MinerYang:
Hi @urpylka,

Thanks for connecting with us! we understood that adding these customized configurations would benefit the using scenarios like you have. However the primary goal of harbor-helm is to deliver our users straightforward way to deploy Harbor into their Kubernetes clusters and less complicity.  From a maintenance perspective, keeping things as simple as possible is often more beneficial. 

e.g:
In your case, to adding a common customized service account while deploy harbor looks reasonable for me. However we should consider more thoroughly, like the upgrade use case(what if users want switch to using a customized service account since we adding it. ), and we should adding more logic for checking the service account existence as long as the compliance of the current `serviceAccountName` config etc..

So if these features you proposed above are more generally been required from our community, we would like to get a next step. Otherwise, I would suggest you to fork our repo and customized at your side to make it more convenient and flexible for you.

Best,
Miner

urpylka:
Hi @MinerYang,

I understand your position, but I want to emphasize that using your chart in its current form is not feasible for companies with established security policies.

Defining secrets in values.yaml is simply unacceptable. The de facto standard today is using solutions like Vault (OpenBao), AWS Secrets Manager, etc.

Furthermore, allowing users to define additional volumes does not contradict the straightforward approach you aim for. This is a widely adopted practice—similar to how your chart already provides the option to define a ServiceAccountName or leave it unset.

Regarding the ServiceAccount, I won’t argue—I agree that my initial proposal wasn’t fully thought through. However, adding this specific change is not crucial as long as you introduce support for extraObjects (as evidenced by another pull request on this topic: #1917).

I am willing to revise my PR (or you can modify it yourselves—I have already provided access) by removing the ServiceAccount creation. However, the ability to define extraObjects, volumes, and volumeMounts is essential. Without these features, it is impossible to securely use your Helm chart.

I also want to stress that adding extraObjects, volumes, and volumeMounts does not introduce unnecessary complexity in terms of maintenance. On the contrary, keeping things flexible in this way can often be more beneficial, as it allows users to add features in the future without changing the Helm chart, while still maintaining simplicity.

urpylka:
Nevertheless, I acknowledge that I may be misunderstanding something. Therefore, I am open to your suggestions on how to use your chart with https://secrets-store-csi-driver.sigs.k8s.io/.

In any case, I look forward to hearing back from you.

PR #2164, opened 2025-04-07
fix: improve helm lookup fallback in case of helm template rendering

khanhngobackend:
To re-PR for a **closed** bot-marked stale one (approved but wrongly assumed for it to be auto-merged): goharbor/harbor-helm#1897

This improve the case of using `helm template .` to statically render the resources. Without this, it might produce error similar to the following even with default `values.yaml`

```
Error: template: harbor/templates/registry/registry-secret.yaml:12:62: executing "harbor/templates/registry/registry-secret.yaml" at <include "harbor.secretKeyHelper" (dict "key" "REGISTRY_HTTP_SECRET" "data" $existingSecret.data)>: error calling include: template: harbor/templates/_helpers.tpl:59:41: executing "harbor.secretKeyHelper" at <.data>: wrong type for value; expected map[string]interface {}; got interface {}
```

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -56,8 +56,8 @@ app: "{{ template "harbor.name" . }}"
 
 {{/* Helper for printing values from existing secrets*/}}
 {{- define "harbor.secretKeyHelper" -}}
-  {{- if and (not (empty .data)) (hasKey .data .key) }}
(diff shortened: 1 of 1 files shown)


sstolbov:
Nice try! What about harbor.redis.pwdfromsecret on 175 raw? Could you please look here also.

PR #2171, opened 2025-04-15
chore: Allow overriding the data path for Redis

EyeCantCU:
Some Redis images use /data as the directory for data instead of /var/lib/redis. Allow overriding this so that other images may be used


MinerYang:
Hi @EyeCantCU ,

Could you provided more context why you want override the data path of redis pod/container?
This volume is defined as `emptyDir` and only created when a pod is assigned.

PR #2175, opened 2025-04-18
feat(gw-api): add support for HTTPRoutes

larivierec:


MinerYang:
Hi @larivierec ,

Thanks for reaching out to us and contribute to Harbor. I understood that using api gateway would benefit in some use case. However we generally use nginx as reverse proxy and some of  ingress-controller tools would take care of the route logic itself. As the api gateway is necessarily needed, I will keep take a look on your PR.

Thanks,
Miner

larivierec:
Yes I can. I'll do that later.

I've set it up here as well: https://github.com/larivierec/home-cluster/blob/main/kubernetes/main/apps/development/harbor/app/route.yaml

The way it's currently handled an option for exposure has to be provided. Therefore, both ingress and the manually route have been added in the above example.

larivierec:
The `README.md` table is a bit all over the place, I re-rendered it and added the 4 expose values with routes and added a bit more clear description in the PR. Let me know if I can do something else

sorrison:
You could simplify the httproute a little bit by grouping the matches to render something like 
```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: harbor
  namespace: harbor
spec:
  parentRefs:
    - name: public
      namespace: infra
  hostnames:
    - registry.mydomain
  rules:
    - matches:
        - path:
            value: /v2/
        - path:
            value: /service/
        - path:
            value: /c/
        - path:
            value: /api/
      backendRefs:
        - name: harbor-core
          port: 80
    - matches:
        - path:
            value: /
      backendRefs:
        - name: harbor-portal
          port: 80
```

This is what I'm currently using in our deployment

larivierec:
Hey there @MinerYang, any update from your end relating to this PR?
Anything you'd like me to add or do?

templates/nginx/secret.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.expose.type "ingress" }}
```
wy65701436, line 1: as for the nginx deployment, it also needs to be disabled in http-route scenario.

wy65701436: as well as the cm of nginx.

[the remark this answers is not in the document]
larivierec: Will do, anything else I should cover?
This will be fixed in my next commit

PR #2178, opened 2025-04-29
Support per-service external Redis instances for Harbor services

srbhklkrn:
## What this PR does

This PR adds support for configuring separate external Redis instances for each Harbor service component (core, jobservice, registry, trivyAdapter, etc.).

Currently, Harbor Helm chart only supports connecting all services to a single Redis instance and separates their data using logical DBs (database index). However, many enterprise-grade Redis offerings (such as managed Redis services and Redis Enterprise) do not support logical databases, and only database index 0 is available.

This limitation prevents Harbor from working as intended when using such external Redis services.

### What's changed

- Added a new `services:` block under `redis.external` in `values.yaml` to configure per-service overrides for:
  - `addr`
  - `username`
  - `password`
  - `existingSecret`
- Extended Helm templates:
  - `addrFor`, `passwordFor`, `credFor` now prioritize `services.<svc>.addr` etc. before falling back to global settings.
- Updated `registry-cm.yaml` and `registry-secret.yaml` to use new per-service Redis credentials.
- Fully backward compatible: if no per-service values are provided, Harbor behaves exactly as before.
- Legacy flat keys like `coreAddr`, `registryAddr`, etc. remain supported for now.

### Why this is needed

When using external Redis services where logical DBs are not available (DB index always 0), each Harbor component must connect to a separate Redis instance to prevent data collisions and ensure Harbor works properly.

This PR allows users to configure Harbor cleanly for such environments without needing complex workarounds.

### Example values.yaml structure

```yaml
redis:
  type: external
  external:
    addr: redis.shared:6379
    username: ""
    password: ""
    existingSecret: ""

    services:
      core:
        addr: redis-core:6379
        username: coreuser
        password: corepass
      jobservice:
        addr: redis-job:6379
        existingSecret: redis-jobsecret
      registry:
        addr: redis-registry:6379
        password: regpass

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -173,21 +173,70 @@ app: "{{ template "harbor.name" . }}"
   {{- end }}
 {{- end -}}
 
-{{- define "harbor.redis.masterSet" -}}
-  {{- with .Values.redis }}
-    {{- ternary .external.sentinelMasterSet "" (contains "+sentinel" (include "harbor.redis.scheme" $)) }}
+{{/* host:port for a particular Harbor service, or fall back to addr */}}
+{{- define "harbor.redis.addrFor" -}}
+  {{- $svc := index . 0 -}}
+  {{- $ctx := index . 1 -}}
+  {{- with $ctx.Values.redis }}
+    {{- if eq .type "internal" }}
+      {{- printf "%s:6379" (include "harbor.redis" $ctx) }}
+    {{- else }}
+      {{- $svcMap := index .external.services $svc | default dict }}
+      {{- if $svcMap.addr }}
+        {{- $svcMap.addr }}
+      {{- else }}
+        {{- $legacyKey := printf "%sAddr" $svc }}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by MinerYang** [comment]
Hi @srbhklkrn ,
Thanks for reaching out to us,
There's a workable way that could help you to solve the issue in this enterprise redis case, modify each of the redis configuration or redis url in the config manifest per components alternatively.
e.g. in `core-cm.yaml`
```
_REDIS_URL_CORE:
_REDIS_URL_REG:
```
We understood this may not convenient enough. however we would like to keep open this PR in the community until collecting this requirement more generally.

Best,
Miner

srbhklkrn:
Thanks @MinerYang appreciate the response.

Just to clarify: while _REDIS_URL_* env overrides work for components like core and jobservice, they don’t apply to registry, which reads Redis settings directly from the rendered ConfigMap. That makes it hard to fully rely on the workaround when using multiple Redis instances.

This PR solves that gap by allowing clean, per-service Redis config without relying on deep overrides or hacks. It's backward-compatible and keeps logic within Helm where users expect it.

PR #2206, opened 2025-06-10
Add support config namespace and timeout in serviceMonitor

Alja9:
In this PR, it support users to be able to custom namespace for serviceMonitor harbor into different namespace.
Example in my case : I have center monitoring in one namespace.

And also, because serviceMonitor already support custom interval, it would be nice to have custom scrapeTimeout.
Example in my case : interval and scrapeTimout must be the same value

### DIFF
--- a/templates/nginx/configmap-https.yaml
+++ b/templates/nginx/configmap-https.yaml
@@ -138,6 +138,8 @@ data:
           proxy_set_header X-Forwarded-Proto $x_forwarded_proto;
           proxy_buffering off;
           proxy_request_buffering off;
+          proxy_send_timeout 900;
+          proxy_read_timeout 900;
         }
 
         location /service/ {


MinerYang:
Hi @Alja9 ,

- From harbor perspective, any resources that beyond harbor namespace should not be  created/managed by harbor. So I guess in your case you need to create the serviceMonitor in your center monitor namespace in place rather than render with harbor helm.
- For the scrapeTimeout,  we will collect more needs from community for the same requirement and consider to add this configurations. Before this I'm afraid you need to fork and customize your helm if needed.

Alja9:
Yes, I see. Thank you for your review.

Alja9:
Hi @MinerYang , I have question, how about adding the nodePort on [exporter service](https://github.com/goharbor/harbor-helm/blob/main/templates/exporter/exporter-svc.yaml) for expose the metrics from namespace isolation (replace custom namespace svc monitor), can we add it ?

Alja9:
I will leave this PR for the costumize timeout feature, and let me know again if there is any feedback regarding the timeout feature, I appreciate any feedback. Thanks.

Issue #2209, opened 2025-06-19
Need to add a configure value for max_job_duration_hours for jobservice


stonezdj:
See: https://github.com/goharbor/harbor/pull/21390/files 

Because harbor.yml already add this configuration `max_job_duration_hours` in 2.13.0, this value should be added to the values.yaml

MinerYang:
Hold until upstream issue resolved: https://github.com/goharbor/harbor/issues/22105

MinerYang:
https://github.com/goharbor/harbor/pull/22116/files
Due to above upstream change, we do not need define MAX_JOB_DURATION_SECONDS explicitly in jobservice env, since jobservice backend would handle the logic by reading reaper.max_job_duration_hours

PR #2211, opened 2025-06-20
[DO-NOT-MERGE]add_jobservice_max_duration

MinerYang:
fix: https://github.com/goharbor/harbor-helm/issues/2209


MinerYang:
due to upstream change, we do not need define `MAX_JOB_DURATION_SECONDS` explicitly in jobservice env, since jobservice backend would handle the logic by reading `reaper.max_job_duration_hours`

Issue #2205, opened 2025-06-06
Missing "spec.ipFamilyPolicy" and "spec.ipFamilies" properties in Harbor services


gsk1985:
Hi Team,
Harbor service template (https://github.com/goharbor/harbor-helm/blob/main/templates/core/core-svc.yaml)  missing the "spec.ipFamilyPolicy" and "spec.ipFamilies". **Its missing for all the harbor services.**

As a result when harbor is deployed in dual stack cluster, We are seeing only IPv6 assigned to the harbor services also ipFamilyPolicy is coming as SingleStack.   **As per kubernetes doc**  (https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-options-on-new-services)

` Service which does not explicitly define .spec.ipFamilyPolicy,  When its created, Kubernetes assigns a cluster IP for the Service from the first configured service-cluster-ip-range and sets the .spec.ipFamilyPolicy to SingleStack.`

```
kubectl describe svc harbor-core 

service/harbor-core  
IP Family Policy: SingleStack  
IP Families: IPv6  
IP: fd10:100:96::d71a  
IPs: fd10:100:96::d71a
```

**is there any plan to introduce ipfamily and ipfamily policies as part of harbor services yaml files?**

MinerYang:
closed by fixed in https://github.com/goharbor/harbor-helm/pull/2212

PR #2212, opened 2025-07-08
feat: Added support for configuring dual-stack services

Xartos:
## Why do we want this

To be able to configure services with dual-stack we need to be able to set the `ipFamilyPolicy` and `ipFamilies` for all services.

Fixes: #2205

## Additional info

This shouldn't change the default behavior at all, only if any of the new values are set.


Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 1.4.0-dev
+version: 1.4.1-dev
```
reasonerjt, line 3: nit: you don't need to update it.  The maintainer will edit it when the release branch is cut.

reasonerjt:
@Xartos 

Thanks for making the update!

Could you please also update `README.md` and undo the change in `Chart.yaml`, and squash the commit?

After that I'll approve and merge this PR.

PR #2214, opened 2025-07-25
Support multiple hosts for core

Jamess11:
Added support for multiple core hosts at the ingress level.
In an air-gapped environment, we needed to add another ingress manually in order to make global docker registries such as docker.io and ghcr.io urls accessible via the habror we deloyed. With this change, it would be much more easy to manage.


reasonerjt:
> we needed to add another ingress manually in order to make global docker registries such as docker.io and ghcr.io urls accessible via the habror we deployed.

I don't understand why Harbor can host images in `ghcr.io`

PR #2223, opened 2025-08-13
update adapter list for replication and proxy-cache

MinerYang:


LordGaav:
Why was this change made? I can't find why Quay is not longer a valid endpoint type, and the `PERMITTED_REGISTRY_TYPES_FOR_PROXY_CACHE` value is not configurable through Helm.

PR #2240, opened 2025-09-10
Add Content-Security-Policy header templating for portal

anders-elastisys:
I wanted to option to add Content-Security-Policy headers for Harbor portal.
This adds templating to be able to configure this, setting the default to an empty string keeping the config the same as before the addition of this templating.

If preferred the templating could be more generic for adding any type of headers, not only Content-Security-Policy.
Let me know if there is another recommended way for configuring CSPs in Harbor, or if it would be preferred to add this as meta tag instead in the html?

### DIFF
--- a/templates/portal/configmap.yaml
+++ b/templates/portal/configmap.yaml
@@ -63,6 +63,9 @@ data:
             }
             location = /index.html {
                 add_header Cache-Control "no-store, no-cache, must-revalidate";
+                {{- with .Values.portal.contentSecurityPolicy }}
+                add_header Content-Security-Policy {{ . }}
+                {{- end }}
             }
         }
     }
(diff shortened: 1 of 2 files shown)


MinerYang:
Hi @anders-elastisys ,

How' s you harbor been deployed? Would add the extra header from your either Nginx reverse proxy or ingress service side instead of adding it to the harbor-portal internal nginx match your expectation?

anders-elastisys:
Hi @MinerYang, Harbor is deployed in a Kubernetes cluster fronted by an Ingress with the Ingress-NGINX controller installed. Adding the header through Ingress is not an option in my case, as I do not want to allow snippet annotations as other users use Ingresses in the same cluster, and configuring a global CSP in the Ingress-NGINX controller is not really an option either as I'd rather be able to configure as strict CSPs as possible without braking other applications backed by the Ingress controller.

anders-elastisys:
@MinerYang who can take a look at this? how does this get assigned a reviewer?

anders-elastisys:
@MinerYang any updates on getting this reviewed?

anders-elastisys:
@MinerYang any possibility getting this merged?

PR #2246, opened 2025-09-19
feat: add support to configure securityContext

Falltrades:
Following the logic of this PR : https://github.com/goharbor/harbor-helm/pull/1695 to avoid error such as the one below when deploying on an Openshift cluster.

```
.spec.securityContext.fsGroup: Invalid value: []int64{10000}: 10000 is not an allowed group, provider restricted-v2: .containers[0].runAsUser: Invalid value: 10000: must be in the ranges: [1000870000, 1000879999]
```

Example test:
```
helm template . --set metrics.enabled=true --set securityContext="" --show-only templates/exporter/
exporter-dpl.yaml 
```

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -36,9 +36,9 @@ spec:
 {{ toYaml .Values.core.podAnnotations | indent 8 }}
 {{- end }}
     spec:
-      securityContext:
-        runAsUser: 10000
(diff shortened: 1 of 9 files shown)


Falltrades:
Hi @MinerYang,
I am currently using 1.16.0 chart version. Indeed `containerSecurityContext` is supported, but it seems we need support for `securitycontext ` as well.

PR #2259, opened 2025-10-17
feat: set log level per container

ricardojdsilva87:
This Pull request allows to enable the logging configuration per Harbor container.

This is useful for example if we want to have the information on the log for the cleanup jobs based on the retention rules on the repositories. The information regarding tags and sizes are only available on the log via UI if the log level is set on the `jobservice` container. To enable this we need to set the `.Values.logLevel` on the `values.yaml` that will enable debug logs for all containers that we might not need.

Enabling in this case the `.Values.jobservice.logLevel` will take precedence over the `.Values.logLevel`. Also leaving the other `logLevel` commented will take the `.Values.logLevel` as default value, making this change backwards compatible

Documentation was also updated

### DIFF
--- a/templates/jobservice/jobservice-cm.yaml
+++ b/templates/jobservice/jobservice-cm.yaml
@@ -25,7 +25,7 @@ data:
     job_loggers:
       {{- if has "file" .Values.jobservice.jobLoggers }}
       - name: "FILE"
-        level: "{{ default .Values.logLevel .Values.jobservice.logLevel }}"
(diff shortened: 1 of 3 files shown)


ricardojdsilva87:
Thank you for your reply @MinerYang,
The issue is that everytime there is an update on the helm chart for a new version for example, those values will be replaced by the one in `{{ .Values.logLevel }`. This change will need to be done everytime manually on the configmap, and if there is another helm apply that will be reverted.
With these changes it gives the possibility to finetune the configurations for each component
Thanks!

PR #2273, opened 2025-11-17
fix(nginx): add namespace to nodePort and loadBalancer services

Rory-Z:


Rory-Z:
> Hi @Rory-Z <
> 
> Is it related to Argo CD diff?

I'm not entirely sure, but I believe it might be effective. I'm using Kustomize with Helm, and the NodePort service isn't being created in the correct namespace. I think Argo CD has the same issue.

Perhaps you could check the details of the Argo CD diff to make sure it.

MinerYang (approved):

PR #2275, opened 2025-11-24
Add monitoring port names to the harbor component deployments.

rikuwe-christian-rieger:
We use  grafana alloy with its pod discovery feature, it is way easier to identify the metrics ports if they are correctly named in the manifests.


MinerYang:
Hi @rikuwe-christian-rieger ,

Thanks for connecting with us.
I am not familiar with pod discovering feature,  when grafana enable pod discovery feature, is the container-port necessarily set?

PR #2289, opened 2025-12-16
Skip creating core/jobservice secrets when using external secrets

eidmantas:
Hello, I noticed that when you move all secrets to an external (non-helm generated) source - the secrets still remain and cause confusion for peers.

I thought about adding a helper function and if everything is moved, then we do not create a secret at all.


eidmantas:
Also now I am thinking, we probably need to deal with annotations as well? Or is cksum of 0 fine? At least local templating worked fine.

cthtrifork:
This is still relevant

cthtrifork:
This is still relevant

PR #2310, opened 2026-02-03
fix: reuse existing secretKey and TLS certs to prevent ArgoCD drift

### LINKED ISSUE #2263: ArgoCD: drift: randAlphaNum
Hey, 
I've deployed harbor with ArgoCD. I receive several notifications, that harbor has been successfully synced, even though nothing has been changed. 

I investigated some time to analyze the issue and found that the chart has some defaults implemented to generate random alpha numeric strings. For example in `core-secret.yaml`.

```yaml
  {{- if not .Values.core.existingSecret }}
  secret: {{ .Values.core.secret | default (include "harbor.secretKeyHelper" (dict "key" "secret" "data" $existingSecret.data)) | default (randAlphaNum 16) | b64enc | quote }}
  {{- end }}
```

Every time ArgoCD renders the helm chart, there is a new generic alphanumeric value. This causes ArgoCD to detect a drift and restore the original state. It should now be clear that this pattern leads to a forced sync. Otherwise, the helm [documentation](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) describes it quite well: 

> [!TIP]
> In the event you always want to roll your deployment, you can use a similar annotation step as above, instead replacing with a random string so it always changes and causes the deployment to roll.
> https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

Basically, I have no problem at all with the approach of using a random string there. However, nowhere does it say that this leads to a drift through ArgoCD. 

I would suggest updating the documentation here and pointing out to other users what this leads to. The current workaround is to ignore all instances where a default is defined by a random string. 

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: harbor
spec:
  project: default
  ignoreDifferences:
  - kind: Secret
    jsonPointers:
    # harbor-core
    - /data/secret
    - /data/CSRF_KEY
    # registry-secret
    - /data/REGISTRY_HTTP_SECRET 
    # jobservice-secret 
    - /data/JOBSERVICE_SECRET
```

Volker

MyoungHaSong:
## Summary

When using ArgoCD or other GitOps tools, the helm template is rendered on each sync. This causes unnecessary secret updates and pod restarts because:

1. `secretKey` did not use the `lookup` function to check for existing values
2. TLS certificates (`tls.key`, `tls.crt`) were always regenerated with `genCA`

## Changes

- **secretKey**: Add lookup fallback to reuse existing value from the cluster
- **tls.key/tls.crt**: Check for existing certificates before generating new ones with `genCA`

## How it works

The fix follows the same pattern already used for `secret` and `CSRF_KEY`:

```yaml
# Before (secretKey)
secretKey: {{ .Values.secretKey | b64enc | quote }}

# After (secretKey)
secretKey: {{ .Values.secretKey | default (include "harbor.secretKeyHelper" ...) | default "not-a-secure-key" | b64enc | quote }}
```

For TLS certificates, we check if both `tls.key` and `tls.crt` exist in the current secret before falling back to `genCA`.

## Testing

- [x] `helm template` renders successfully
- [x] Fresh install: generates new secrets as expected
- [x] Existing install: reuses existing values (no drift)

Closes #2263

### DIFF
--- a/templates/core/core-secret.yaml
+++ b/templates/core/core-secret.yaml
@@ -9,15 +9,22 @@ metadata:
 type: Opaque
 data:
   {{- if not .Values.existingSecretSecretKey }}
-  secretKey: {{ .Values.secretKey | b64enc | quote }}
+  secretKey: {{ .Values.secretKey | default (include "harbor.secretKeyHelper" (dict "key" "secretKey" "data" $existingSecret.data)) | default "not-a-secure-key" | b64enc | quote }}
   {{- end }}
   {{- if not .Values.core.existingSecret }}
   secret: {{ .Values.core.secret | default (include "harbor.secretKeyHelper" (dict "key" "secret" "data" $existingSecret.data)) | default (randAlphaNum 16) | b64enc | quote }}
   {{- end }}
   {{- if not .Values.core.secretName }}
+  {{- $existingKey := include "harbor.secretKeyHelper" (dict "key" "tls.key" "data" $existingSecret.data) }}
+  {{- $existingCert := include "harbor.secretKeyHelper" (dict "key" "tls.crt" "data" $existingSecret.data) }}
+  {{- if and $existingKey $existingCert }}
+  tls.key: {{ .Values.core.tokenKey | default $existingKey | b64enc | quote }}
+  tls.crt: {{ .Values.core.tokenCert | default $existingCert | b64enc | quote }}
+  {{- else }}
   {{- $ca := genCA "harbor-token-ca" 365 }}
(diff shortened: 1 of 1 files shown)


alessio-pascolini:
Please keep it open

igorrudyk:
Please merge it

alessio-pascolini:
Unstale

WoahitsJoe:
@MinerYang Any chance on getting this merged? It's causing replicaset revision spam

WoahitsJoe:
@stonezdj

PR #2319, opened 2026-02-27
feat: added support hostAliases

eXist-FraGGer:
# Added hostAliases support to all Harbor Helm Deployments:
- core
- portal
- jobservice
- registry
- nginx
- exporter

This enables consistent host alias customization across all Harbor components without breaking existing installs.
Default behavior is unchanged because `hostAliases` is empty by default.

# Changes
1. Updated Deployment templates to render spec.template.spec.hostAliases conditionally from values:
```yaml
{{- with .Values.<component>.hostAliases }}
hostAliases:
{{ toYaml . | indent 8 }}
{{- end }}
```
2. Added default values in values.yaml:
```yaml
nginx:
  hostAliases: []
portal:
  hostAliases: []
core:
  hostAliases: []
jobservice:
  hostAliases: []
registry:
  hostAliases: []
exporter:
  hostAliases: []
```

# Validation
- `helm template harbor .` passes with default values.
- `helm template harbor . -f /tmp/harbor-hostaliases-all-values.yaml` passes with sample `hostAliases` overrides.
- Verified `hostAliases` appears in rendered Deployment manifests when configured.

### DIFF
--- a/templates/core/core-dpl.yaml
+++ b/templates/core/core-dpl.yaml
@@ -58,6 +58,10 @@ spec:
             component: core
 {{- end }}
 {{- end }}
+      {{- with .Values.core.hostAliases }}
+      hostAliases:
+{{ toYaml . | indent 8 }}
+      {{- end }}
       {{- with .Values.core.initContainers }}
       initContainers:
         {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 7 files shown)


eXist-FraGGer:
Hi @MinerYang, could you please re-open this PR?

PR #2322, opened 2026-03-04
Expose REGISTRY_HTTP_CLIENT_TIMEOUT for jobservice

### LINKED ISSUE #2313: Expose REGISTRY_HTTP_CLIENT_TIMEOUT via Helm Values for Registry Deployment
### Context
Our current deployments operate in restricted, high-latency, and low-bandwidth network environments. During operations such as proxy-caching or large image pulls from upstream registries, we are observing frequent failures.

### Problem:
The Harbor Registry component uses a hardcoded default of 30 seconds for its internal HTTP client timeout. In our specific network conditions, this window is insufficient, leading to context canceled or timeout errors during the proxy-cache process.

### Proposed Solution:
Introduce a configurable parameter in the values.yaml to allow for the dynamic setting of the REGISTRY_HTTP_CLIENT_TIMEOUT environment variable within the Registry deployment template.

Proposed Key: registry.httpClientTimeout

Default Value: 30

Implementation: Inject the value into the registry deployment container spec as an environment variable.

MinerYang:
Fix: https://github.com/goharbor/harbor-helm/issues/2313


stonezdj (approved):

neevnuv:
@MinerYang, accourding to this - https://github.com/goharbor/harbor/issues/22393#issuecomment-3345998709, the REGISTRY_HTTP_CLIENT_TIMEOUT chould be applied to the harbor-core config. Should we add this option additionally to the core configMap?

PR #2359, opened 2026-05-02
feat(nginx): add loadBalancerClass support to LB svc

mlguerrero12:
Add spec.loadBalancerClass rendering to the nginx Service template when expose.type=loadBalancer. The field is omitted when unset, preserving backward compatibility for existing deployments.


mlguerrero12:
@MinerYang PTAL. Thanks!

MinerYang:
> @MinerYang PTAL. Thanks!

Hi @mlguerrero12 ,

Sorry for the late response. Could you also help to write a UT for this specific case?
e.g. https://github.com/goharbor/harbor-helm/tree/main/test/unittest

Thanks^
Miner

PR #2370, opened 2026-05-12
Digest pinning support for Trivy image.

Vinnie161:
## Summary

This change adds optional digest support for the Harbor Trivy adapter image.

Currently the Trivy StatefulSet always renders the image as:

```
  <repository>:<tag>
```

This makes it impossible to use digest-based image references such as:

```
  goharbor/trivy-adapter-photon@sha256:<digest>
```

The new `trivy.image.digest` value is optional and backward compatible:
- when `trivy.image.digest` is empty, the chart keeps rendering `<repository>:<tag>`
- when `trivy.image.digest` is set, the chart renders `<repository>@<digest>`

This allows users to pin the Trivy adapter image by immutable digest while preserving the existing tag-based behavior by default.

### DIFF
--- a/templates/trivy/trivy-sts.yaml
+++ b/templates/trivy/trivy-sts.yaml
@@ -63,7 +63,7 @@ spec:
       {{- end }}
       containers:
         - name: trivy
-          image: {{ .Values.trivy.image.repository }}:{{ .Values.trivy.image.tag }}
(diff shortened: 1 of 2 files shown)


MinerYang:
Let's discuss more in this thread
https://github.com/goharbor/harbor-helm/issues/2372#issuecomment-4474756285

MinerYang:
Hi @Vinnie161 ,

Sorry for the late response. Would you like to continue working on this PR and add a UT and update README for this configuration?

Thanks,
Miner

Vinnie161:
Hi, @MinerYang!
Sorry, I'm overloaded right now. I'll try to get back to this PR when I have some time.

PR #2382, opened 2026-06-29
Support existing secret for the full external DB connection

### LINKED ISSUE #2381: Support sourcing the full external DB connection from an existing secret
### Problem

When using an external database (`database.type: external`), only the **password** can be supplied via `database.external.existingSecret` (and the key is hardcoded to `password`). The remaining connection parameters — `host`, `port`, `username`, `coreDatabase` — must be set as **plaintext values** in `values.yaml`.

This is awkward when the connection details are themselves sensitive or are produced by another component (e.g. an operator/cloud provisioner that emits a single connection secret), and it prevents a parent/umbrella chart from injecting a fully secret-driven connection.

### Proposal

Make `database.external.existingSecret` an object so the **entire** connection can come from one existing secret, with configurable key names:

```yaml
database:
  external:
    existingSecret:
      enabled: false
      name: "my-external-secret-name"
      keys:
        host: host
        port: port
        username: username
        password: password
        coreDatabase: dbName
```

When `enabled: true`, core (deployment + pre-upgrade job) and exporter read the connection from the referenced secret, and the configmap/secret templates skip emitting the plaintext values. When `enabled: false`, behaviour is unchanged (backward compatible).

### Notes

This revives the approach from the previously closed #1783. I have a PR ready against `main`.

tada5hi:
Closes #2381.

Revives the change from #1783 (closed unmerged), rebased on current `main`.

### What

Today, with an external database, only the password can be read from `database.external.existingSecret` (with the key hardcoded to `password`); `host`, `port`, `username` and `coreDatabase` must be plaintext values. This PR turns `existingSecret` into an object so the **whole** connection can come from one secret, with configurable key names:

```yaml
database:
  external:
    existingSecret:
      enabled: false
      name: "my-external-secret-name"
      keys:
        host: host
        port: port
        username: username
        password: password
        coreDatabase: dbName
```

- When `enabled: true`: `core` (deployment + pre-upgrade job) and `exporter` source `POSTGRESQL_HOST/PORT/USERNAME/PASSWORD/DATABASE` from the referenced secret via `secretKeyRef`, and `core-cm` / `core-secret` / `exporter-cm-env` / `exporter-secret` skip emitting the plaintext values.
- When `enabled: false`: behaviour is unchanged (backward compatible).

### Testing

- `helm lint` → 0 failed.
- `helm template` with `database.type=external` and `database.external.existingSecret.enabled=true` renders all five connection variables from the referenced secret on both core and exporter; with it disabled, the rendered output matches `main`.

### DIFF
--- a/templates/core/core-cm.yaml
+++ b/templates/core/core-cm.yaml
@@ -15,10 +15,12 @@ data:
     httpport = {{ ternary "8443" "8080" .Values.internalTLS.enabled }}
   PORT: "{{ ternary "8443" "8080" .Values.internalTLS.enabled }}"
   DATABASE_TYPE: "postgresql"
+  {{- if not .Values.database.external.existingSecret.enabled }}
   POSTGRESQL_HOST: "{{ template "harbor.database.host" . }}"
   POSTGRESQL_PORT: "{{ template "harbor.database.port" . }}"
   POSTGRESQL_USERNAME: "{{ template "harbor.database.username" . }}"
   POSTGRESQL_DATABASE: "{{ template "harbor.database.coreDatabase" . }}"
+  {{- end }}
   POSTGRESQL_SSLMODE: "{{ template "harbor.database.sslmode" . }}"
   POSTGRESQL_MAX_IDLE_CONNS: "{{ .Values.database.maxIdleConns }}"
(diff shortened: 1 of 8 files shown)


tada5hi:
Still relevant

PR #2384, opened 2026-06-29
Render external database connection values with tpl

### LINKED ISSUE #2383: Render external database connection values with tpl (release-name-based host)
### Problem

The external database connection values — `database.external.host`, `port`, `username`, `coreDatabase`, `sslmode` — are emitted verbatim (the `harbor.database.*` helpers return the raw value). They are **not** rendered through `tpl`, so a value like:

```yaml
database:
  external:
    host: "{{ .Release.Name }}-postgresql-primary"
```

ends up as the literal string `{{ .Release.Name }}-postgresql-primary` instead of being interpolated.

This makes it impossible for a parent/umbrella chart to point Harbor at a **release-scoped** database service without hardcoding the release name — which in turn blocks installing the umbrella chart multiple times in one namespace. Several other charts (e.g. the Bitnami Harbor chart) render these values with `tpl`, so this is a behaviour gap when migrating.

### Proposal

Render the external connection values through `tpl`:

```
{{- tpl (.Values.database.external.host | toString) $ -}}
```

A value with no template syntax is returned unchanged, so the change is fully backward compatible. The password is left untouched.

I have a small PR ready against `main`.

tada5hi:
Closes #2383.

### What
The external DB connection values (`database.external.host`, `port`, `username`, `coreDatabase`, `sslmode`) are emitted verbatim by the `harbor.database.*` helpers, so they cannot contain template expressions. This wraps the external branch of those helpers in `tpl`:

```
{{- tpl (.Values.database.external.host | toString) $ -}}
```

so values may reference the release context, e.g.:

```yaml
database:
  external:
    host: "{{ .Release.Name }}-postgresql-primary"
```

The `password` helper is intentionally left untouched.

### Why
Lets a parent/umbrella chart point Harbor at a release-scoped database service without hardcoding the release name — enabling multiple installs in one namespace.

### Compatibility
A value with no template syntax renders unchanged, so this is fully backward compatible.

### Testing
- `helm lint` → 0 failed.
- `helm template` with `database.type=external`:
  - `host: "{{ .Release.Name }}-postgresql-primary"` → `POSTGRESQL_HOST: "myrel-postgresql-primary"`
  - plain `host: "db.example.com"` → `POSTGRESQL_HOST: "db.example.com"` (unchanged)

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -89,7 +89,7 @@ app: "{{ template "harbor.name" . }}"
   {{- if eq .Values.database.type "internal" -}}
     {{- template "harbor.database" . }}
   {{- else -}}
-    {{- .Values.database.external.host -}}
+    {{- tpl (.Values.database.external.host | toString) $ -}}
   {{- end -}}
 {{- end -}}
 
@@ -97,7 +97,7 @@ app: "{{ template "harbor.name" . }}"
   {{- if eq .Values.database.type "internal" -}}
     {{- printf "%s" "5432" -}}
   {{- else -}}
-    {{- .Values.database.external.port -}}
+    {{- tpl (.Values.database.external.port | toString) $ -}}
   {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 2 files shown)


outofrange:
Would live to see this being merged! #notstale

PR #2386, opened 2026-07-03
fix: validate empty route parentRefs and omit empty hostnames

### LINKED ISSUE #2385: expose.type=route: HTTPRoute template fails to render with chart's own default (empty) parentRefs/hosts
## Describe the bug

When `expose.type: route` (Gateway API HTTPRoute) is used, `helm template` / `helm install` / `helm upgrade` fails with a YAML parse error. This reproduces with the chart's own **unmodified default values** — `expose.route.parentRefs` and `expose.route.hosts` both default to `[]` in `values.yaml` — so `expose.type: route` is broken out of the box, with no user override required.

## Chart version

- harbor-helm: `1.19.1`
- Also confirmed identical, unpatched code on `main` as of this writing, so it isn't fixed in a later release either.
- Helm: `version.BuildInfo{Version:"v4.2.0", GitCommit:"06468084e85c244c712834933d25ea232a4c2093", GitTreeState:"clean", GoVersion:"go1.26.3", KubeClientVersion:"v1.36"}`

## To Reproduce

**Case 1 — chart defaults, nothing overridden:**

```yaml
expose:
  type: route
  tls:
    enabled: false
```

```
helm template harbor harbor/harbor -f values.yaml --version 1.19.1
```

**Case 2 — parentRefs supplied, hosts left at default `[]`:**

```yaml
expose:
  type: route
  tls:
    enabled: false
  route:
    parentRefs:
      - name: some-gateway
        namespace: gateway-ns
        sectionName: https
        kind: Gateway
        group: gateway.networking.k8s.io
```

## Expected behavior

Chart renders valid YAML, or fails with a clear validation message if `parentRefs`/`hosts` are required for this expose mode.

## Actual behavior

Case 1:
```
Error: YAML parse error on harbor/templates/gateway-apis/route.yaml: error converting YAML to JSON: yaml: line 9: could not find expected ':'
```

Case 2 (error moves further down, to the `hostnames` block):
```
Error: YAML parse error on harbor/templates/gateway-apis/route.yaml: error converting YAML to JSON: yaml: line 15: could not find expected ':'
```

## Root cause

`templates/gateway-apis/route.yaml`, lines 25–28:

```yaml
spec:
  parentRefs:
  {{- toYaml $route.parentRefs | nindent 2 }}
  hostnames:
  {{- toYaml $route.hosts | nindent 2 }}
```

When `$route.parentRefs` or `$route.hosts` is an empty list, `toYaml` returns the flow-style literal `[]` instead of nothing, and `nindent 2` places that `[]` on its own line at the **same indentation as the parent key**:

```yaml
  hostnames:
  []
  rules:
```

A value on the line after a key, at the same (or lesser) indentation as that key, isn't valid block-style YAML — hence "could not find expected ':'". This will reproduce for any empty list passed through this `toYaml | nindent` pattern, not just these two specific fields.

## Suggested fix

Guard each block the same way `labels`/`annotations` already are elsewhere in this same file, falling back to inline flow-style `[]` when empty:

```diff
 spec:
-  parentRefs:
-  {{- toYaml $route.parentRefs | nindent 2 }}
-  hostnames:
-  {{- toYaml $route.hosts | nindent 2 }}
+{{- if $route.parentRefs }}
+  parentRefs:
+  {{- toYaml $route.parentRefs | nindent 2 }}
+{{- else }}
+  parentRefs: []
+{{- end }}
+{{- if $route.hosts }}
+  hostnames:
+  {{- toYaml $route.hosts | nindent 2 }}
+{{- else }}
+  hostnames: []
+{{- end }}
   rules:
```

Separately, since an HTTPRoute with empty `parentRefs` can't attach to any Gateway, a `required`/fail-fast check on `expose.route.parentRefs` when `expose.type: route` might be a worthwhile addition on top of the parsing fix.

## Additional context

- Workaround in the meantime: explicitly set at least one entry in both `expose.route.parentRefs` and `expose.route.hosts`.
- File: https://github.com/goharbor/harbor-helm/blob/v1.19.1/templates/gateway-apis/route.yaml#L24-L29

somaz94:
### Problem

When `expose.type: route` is used with the chart's own default values (`expose.route.parentRefs: []` and `expose.route.hosts: []`), `helm template` fails with:

    Error: YAML parse error on harbor/templates/gateway-apis/route.yaml: error converting YAML to JSON: yaml: line 9: could not find expected ':'

`toYaml` on an empty list returns the flow literal `[]`, and `nindent 2` put it at the same indentation as the parent `parentRefs:` / `hostnames:` key, so it parsed as a sibling node. `expose.type: route` was broken out of the box, with a cryptic error and no user override required.

Fixes #2385

### Fix

`parentRefs` and `hostnames` are treated differently, matching their roles in the Gateway API HTTPRoute spec:

- **`parentRefs`** — an HTTPRoute with no `parentRefs` attaches to no Gateway and cannot serve traffic, so it is effectively required in route mode. When it is empty the template now `fail`s with a clear message instead of emitting invalid YAML.
- **`hostnames`** — genuinely optional per the spec (omitting it matches all hostnames), so it is wrapped in a `{{- with }}` guard and omitted when the list is empty, rendered at `nindent 4` when present.

### Validation

- `helm template --set expose.type=route,expose.tls.enabled=false .` now fails with a clear `expose.route.parentRefs must be set …` message (previously a cryptic YAML parse error).
- With `parentRefs` set and `hosts` empty, `parentRefs` renders and `hostnames` is omitted.
- `helm unittest -f 'test/unittest/gateway-apis/*.yaml' .` passes (updated the empty-defaults case to assert the clear failure; added the parentRefs-set/hosts-empty case).
- `helm lint .` passes.

### DIFF
--- a/templates/gateway-apis/route.yaml
+++ b/templates/gateway-apis/route.yaml
@@ -22,10 +22,15 @@ metadata:
 {{ toYaml $route.annotations | indent 4 }}
 {{- end }}
 spec:
+  {{- if not $route.parentRefs }}
+    {{- fail "expose.route.parentRefs must be set when expose.type is \"route\": an HTTPRoute with no parentRefs attaches to no Gateway and cannot serve traffic." }}
+  {{- end }}
   parentRefs:
-  {{- toYaml $route.parentRefs | nindent 2 }}
+    {{- toYaml $route.parentRefs | nindent 4 }}
+  {{- with $route.hosts }}
   hostnames:
-  {{- toYaml $route.hosts | nindent 2 }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   rules:
   - matches:
     - path:
--- a/test/unittest/gateway-apis/route_test.yaml
+++ b/test/unittest/gateway-apis/route_test.yaml
@@ -94,3 +94,30 @@ tests:
       - equal:
           path: spec.rules[1].matches[0].path.value
           value: /
+
+  - it: Route fails with a clear message when parentRefs is empty
+    set:
+      expose:
+        type: route
+    template: templates/gateway-apis/route.yaml
+    asserts:
+      - failedTemplate:
+          errorMessage: 'expose.route.parentRefs must be set when expose.type is "route": an HTTPRoute with no parentRefs attaches to no Gateway and cannot serve traffic.'
+
+  - it: Route renders with parentRefs set but hosts left empty
+    set:
+      expose:
+        type: route
+        route:
+          parentRefs:
+            - name: test-parent
+              namespace: test-namespace
+    template: templates/gateway-apis/route.yaml
+    asserts:
+      - hasDocuments:
+          count: 1
+      - equal:
+          path: spec.parentRefs[0].name
+          value: test-parent
+      - notExists:
+          path: spec.hostnames
--- a/values.yaml
+++ b/values.yaml
@@ -51,7 +51,8 @@ expose:
   route:
     labels: {}
     annotations: {}
-    # References to the parent gateways
+    # References to the parent gateways. Required when expose.type is "route":
+    # an HTTPRoute with no parentRefs attaches to no Gateway and cannot serve traffic.
     parentRefs: []
       # - name: envoy-internal
       #   namespace: networking


MinerYang:
Hi @somaz94 ,

the `parentRef` define which Gateways this Route wants to be attached to and normally we regard it as not omitted.
Could you elaborate more your use case and ping me for a review once it's ready.

Thanks,
Miner

somaz94:
Hi @MinerYang, thanks for the review!

You're right that `parentRefs` shouldn't be omitted — an HTTPRoute with no `parentRefs` attaches to no Gateway and can't serve traffic. My original patch was too lenient there.

To clarify the use case: this started as a pure rendering bug (#2385). The chart ships `expose.route.parentRefs: []` and `hosts: []` as its **own defaults**, so `helm template --set expose.type=route` fails out of the box with a cryptic `could not find expected ':'` YAML parse error — `toYaml` on an empty list emits `[]` at the wrong indent. There's no user "use case" for empty `parentRefs`; the problem is just that the default value produces an unreadable error instead of a clear one.

I've reworked the PR to match your view:

- **`parentRefs`**: no longer omitted. If it's empty under `expose.type: route`, the template now `fail`s with a clear message instead of emitting invalid YAML.
- **`hostnames`**: kept optional via `{{- with }}`, since it's genuinely optional in the Gateway API HTTPRoute spec (omitting it matches all hostnames).

`helm unittest` / `helm lint` pass. Flipping to ready for review now.

Thanks!

somaz94:
Hi @MinerYang — thanks, and I think we're actually in full agreement: this PR does **not** make `parentRefs` optional/omittable. It keeps it required and just replaces a broken out-of-the-box experience with a clear error.

**The use case is the chart's own defaults** (issue #2385): with `expose.type: route` and the shipped defaults `expose.route.parentRefs: []` and `expose.route.hosts: []`, `helm template` fails with a cryptic YAML parse error:

```
Error: YAML parse error on harbor/templates/gateway-apis/route.yaml: error converting YAML to JSON: yaml: line 9: could not find expected ':'
```

`toYaml []` renders the flow literal `[]`, and at `nindent 2` it lands at the same indentation as the parent `parentRefs:` key, so it parses as a sibling and the document is invalid — route mode is unusable until the user happens to set a value, with no hint why.

The fix keeps the Gateway API semantics you describe:
- **`parentRefs`** — since an HTTPRoute with no `parentRefs` attaches to no Gateway and can't serve traffic, it stays *required*. When empty the template now `fail`s with `expose.route.parentRefs must be set when expose.type is "route" …` instead of the cryptic parse error. (Verified it renders that message on the defaults.)
- **`hostnames`** — genuinely optional per the spec (omitting matches all hostnames), so it's wrapped in `{{ with }}` and omitted when empty, which also removes the same invalid-YAML issue for that field.

`helm unittest` (6/6) and `helm lint` pass. So the behavior you'd expect — parentRefs required — is preserved; this only turns a silent broken default into an actionable message. Ready for review whenever you have a moment 🙏

somaz94:
@MinerYang I see the `needs/follow-up` label — could you clarify what you'd like from me here? I want to make sure I address the right thing rather than guess.

To restate the scope: this isn't about making `parentRefs` optional. With the chart's own default values the rendered HTTPRoute is invalid (empty `parentRefs`, and an empty `hostnames` entry), which is the #2385 report. The change validates `parentRefs` and omits `hostnames` when empty, so the default render is valid.

If the follow-up is about something else — more unit-test coverage, narrowing the scope to just the `hostnames` omission, or a different approach entirely — happy to do it, just let me know which. Also note the Lint/Unit/Integration workflows are still pending approval to run on this fork PR, so CI hasn't actually exercised the added unittest yet.

vectinx:
Hi everyone!
I’m experiencing the same issue. 
Is there any update on its current status?
Thanks!

somaz94:
@MinerYang — @vectinx hit this on the shipped defaults too, so it isn't just me. Could you approve the workflows on this PR? Lint, Unit test and Integration test have never run on it, so the unittest I added has never actually been exercised in CI. And if `needs/follow-up` is still waiting on something from me, could you say what — my 28 Jul question about it is still open and I'd rather fix the right thing than guess.

MinerYang (approved):
lgtm

PR #2388, opened 2026-07-16
Add UT for set proxy env for registry

MinerYang:
Add UT that user append `registry` to the proxy list if you want to apply proxy settings to the registry component, for example when using S3 with `disableredirect: true` and need a proxy


wy65701436 (approved):
lgtm

stonezdj (approved):

PR #2391, opened 2026-07-21
feat(metrics): add optional Redis exporter sidecar for internal Valkey

FutureMatt:
Adds an optional `oliver006/redis_exporter` sidecar to the internal Redis/Valkey StatefulSet so operators can scrape Valkey metrics with Prometheus.

- Gated by `metrics.enabled` and `metrics.redis.enabled` (only when `redis.type` is `internal`)
- Exposes the metrics port on the Redis Service using `harbor.metricsPortName`, so the existing ServiceMonitor picks it up
- Configurable image, port, path, resources, and `extraEnvVars` under `metrics.redis`
- Includes helm-unittest coverage for the StatefulSet and Service


FutureMatt:
@MinerYang any chance of a review on this?

PR #2393, opened 2026-07-23
feat: allow trivy github token to be read from existing secret

beschoenen:
This pull requests allows the trivy github token to be retrieved from an existing secret. This eases rotation when using tools like external secrets


beschoenen:
Bump

PR #2412, opened 2026-09-15
docs: document deterministic GitOps secrets

WoahitsJoe:
## Summary

- document the complete existing-Secret configuration needed for deterministic Argo CD and Flux rendering
- cover Harbor internal credentials, token signing certificates, registry htpasswd, public TLS, and internal TLS
- add a CI regression that compares two independent offline renders byte-for-byte

Helm `lookup` cannot reuse live Secret data when a GitOps controller renders the chart without Kubernetes API access. This change documents the chart's existing deterministic path instead of adding a duplicate values API or relying on ignored differences.

## Validation

- `helm lint .`
- all 119 existing Helm unit tests pass with the repository's pinned Helm unittest v0.4.4
- two independent offline renders using `test/gitops/values.yaml` are identical
- Harbor chart 1.19.2 installed successfully in a disposable Kind cluster with these Secret references
- an unchanged Helm upgrade preserved both external Secret resource versions, every Harbor pod UID, Deployment generations, and Secret checksum annotations

Refs #2263.


WoahitsJoe:
Closing this for now while we complete and validate the deployment-side work before proposing anything upstream.

